PR #12, opened 2020-08-03
Feature - Implement the `load` method on RemoteFeedLoader

thiagoramos23:
## Why is this change necessary?

This PR, if accepted, will implement the code to make RemoteFeedLoader works.

## How does it address the issue?

It was created the code to make possible RemoteFeedLoader use the HttpClient, receive the data and response and parse the result to the appropriate domain model.


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,20 @@ public final class RemoteFeedLoader: FeedLoader {
...
+            case .failure(_):
+                completion(.failure(Error.connectivity))
+                break
+            case .success((let data, let response)):
+                guard response.statusCode == 200,
+                      let _ = try? JSONSerialization.jsonObject(with: data) else {
+                    return completion(.failure(Error.invalidData))
+                }
+                
+                return completion(FeedImageMapper.map(from: data, response: response))
```
caiozullo, line 34: Looks like this logic is duplicated - it's doing the same in the `FeedImageMapper`. So you could remove it.

```suggestion
                completion(FeedImageMapper.map(from: data, response: response))
```

PR #14, opened 2020-08-05
Feed API challenge - Bill Wang

chihyinwang:


FeedAPIChallenge/Feed Feature/FeedImage.swift
```diff
@@ -17,3 +17,12 @@ public struct FeedImage: Hashable {
...
 }
+
+extension FeedImage: Decodable {
+    private enum CodingKeys: String, CodingKey {
+        case id = "image_id"
+        case description = "image_desc"
+        case location = "image_loc"
+        case url = "image_url"
+    }
+}
```
mapostolakis, line 28: Those decodable `CodingKeys` are API-specific implementation details. So ideally, they should be defined in the API module. 

By making the `FeedImage` conform to `Decodable` with those `CodingKeys` in the core 'Feed Feature module' will couple the core module with the API implementation details.

Since you've already defined an API-specific `Item` model, you don't need this extension (it's not in use - so I believe you forgot to remove it). You can remove it to prevent low-level API details leaking into the core module.

```suggestion
```

chihyinwang: Yes, indeed. I did forget to remove this. Thank you!

mapostolakis: Nice one!

PR #17, opened 2020-08-10
The Feed API Challenge - Prudhvi Gadiraju

prufect:


FeedAPIChallenge/Feed Feature/FeedImage.swift
```diff
@@ -17,3 +17,5 @@ public struct FeedImage: Hashable {
 		self.url = url
 	}
 }
+
+extension FeedImage: Decodable {}
```
mapostolakis, line 21: This extension is not used since you're using the Item model in the Feed API module to decode the JSON. You can remove it.

mapostolakis:
Hi @prufect,

I left one comment regarding an unused extension that can be removed. Please, feel free to fix it and push a new commit.

Great job! ✅

PR #24, opened 2020-09-02
Complete Feed API Challenge

cocoataster:


FeedAPIChallenge/Feed Feature/FeedImage.swift
```diff
@@ -17,3 +17,12 @@ public struct FeedImage: Hashable {
...
 }
+
+extension FeedImage: Decodable {
+    enum CodingKeys: String, CodingKey {
+        case id = "image_id"
+        case description = "image_desc"
+        case location = "image_loc"
+        case url = "image_url"
+    }
+}
```
mapostolakis, line 28: Those decodable CodingKeys are API-specific implementation details. So ideally, they should be defined in the Feed API module (in an API-specific data model, for example).

Otherwise, it's leaking lower-level API details into the higher-level FeedImage model.

mapostolakis (commented):
Hi @cocoataster,

I left one comment regarding low-level API-specific details that should be removed from the higher-level FeedImage data model.

Please, feel free to fix it and push a new commit.

Great job! ✅

PR #31, opened 2020-09-12
iOS Lead essentials Feed API challenge

aaronmhr:


FeedAPIChallenge/Feed API/HTTPClient.swift
```diff
@@ -5,7 +5,7 @@
 import Foundation
 
 public protocol HTTPClient {
-	typealias Result = Swift.Result<(Data, HTTPURLResponse), Error>
+    typealias Result = Swift.Result<(data: Data, response: HTTPURLResponse), Error>
```
caiozullo, line 8: Changing the interface of this type can introduce breaking changes in clients and other implementations.  

You can revert it back to the original interface to prevent issues and deal with the names on the implementation side.

caiozullo (commented):
Hi @aaronmhr,

I left just a comment regarding a breaking change in the HTTPClient protocol. Feel free to fix and push it.

Great job! ✅

PR #41, opened 2020-10-09
Erich Flock - Feed API Challenge

erichflock:


FeedAPIChallenge/Feed API/ItemsMapper.swift
```diff
@@ -0,0 +1,44 @@
...
+        let image_desc: String?
+        let image_loc: String?
+        let image_url: URL
+
+        var item: FeedImage {
+            return FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
+        }
+    }
+    
+    func mapJsonResponse(jsonResponse: Data, completion: @escaping ([FeedImage]) -> Void) {
```
caiozullo, line 34: This operation is synchronous, so this method could return the result instead of completing via a closure. It could also be a static method.

```
func mapJsonResponse(jsonResponse: Data, completion: @escaping ([FeedImage]) -> Void)
```

Becomes

```
static func mapJsonResponse(jsonResponse: Data) -> [FeedImage]
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -20,5 +20,34 @@ public final class RemoteFeedLoader: FeedLoader {
...
+                ItemsMapper().mapJsonResponse(jsonResponse: data, completion: { feedImages in
+                    completion(.success(feedImages))
+                })
+            }
+        })
     }
+    
+    private func isValidJson(data: Data) -> Bool {
+        
+        if let _ = try? JSONSerialization.jsonObject(with: data, options: []) {
```
caiozullo, line 46: You can use `JSONSerialization.isValidJSONObject(data)` instead. Docs:

https://developer.apple.com/documentation/foundation/jsonserialization/1418461-isvalidjsonobject

Or you could also leave the JSON validation to the ItemsMapper - so you don't need to deserialize the JSON object twice. ✅

FeedAPIChallenge/Feed Feature/FeedImage.swift
```diff
@@ -4,7 +4,7 @@
 
 import Foundation
 
-public struct FeedImage: Hashable {
+public struct FeedImage: Hashable, Decodable {
```
caiozullo, line 7: The `FeedImage` doesn't need to conform to `Decodable` since the private `Item` is used for decoding the images.

caiozullo (commented):
Hi @erichflock,

I left a few comments that can help you simplify the code. Feel free to push changes.

Great job! ✅

FeedAPIChallenge/Feed API/ItemsMapper.swift
erichflock: What is the advantage of setting the function mapJsonResponse to static? I am always confused when I should use a static method or not, since static functions or static classes are always loaded into memory.

erichflock:
Hi @caiozullo ! Thanks for the feedback. I have updated the PR.

caiozullo:
Nice one!

FeedAPIChallenge/Feed API/ItemsMapper.swift
caiozullo: Hi @erichflock, you can ask questions on the challenge page - also use the search in the academy to find threads about this topic! Thank you.

erichflock: I've found a thread with the answer, thanks!

PR #57, opened 2020-10-25
Adrian Szymanowski - Feed API Challenge

Adrimi:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,49 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 	
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
+        client.get(from: url) { [weak self] result in
+            guard let self = self else { return }
+            self.mapCompletionResult(result, completion: completion)
```
caiozullo, line 24: You can eliminate the `guard` with `self?` - the result is the same with one less line of code.

```suggestion
            self?.mapCompletionResult(result, completion: completion)
```

FeedAPIChallenge/Feed Feature/FeedImage.swift
```diff
@@ -17,3 +17,5 @@ public struct FeedImage: Hashable {
 		self.url = url
 	}
 }
+
+extension FeedImage: Decodable {}
```
caiozullo, line 21: The `FeedImage` doesn't need to conform to `Decodable` since the `Item` struct is used for decoding. You can remove this line.

caiozullo (commented):
Hi @Adrimi,

I left two suggestions. Feel free to push changes.

Great job! ✅

PR #75, opened 2020-11-24
Jean Mouton - Feed API Challenge

jeanm-dev:
Added example implementation for `RemoteFeedLoader` to load `FeedImage` items and decode the JSON data to `FeedImage` items.

### DIFF
--- a/FeedAPIChallenge/Feed Feature/FeedImage.swift
+++ b/FeedAPIChallenge/Feed Feature/FeedImage.swift
@@ -17,3 +17,12 @@ public struct FeedImage: Hashable {
 		self.url = url
 	}
 }
+
+extension FeedImage: Decodable {
+    private enum CodingKeys: String, CodingKey {
+        case id = "image_id"
+        case description = "image_desc"
+        case location = "image_loc"
+        case url = "image_url"
+    }
+}
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -20,8 +20,11 @@ public final class RemoteFeedLoader: FeedLoader {
 	
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
         client.get(from: url) { result in
-            if case .failure = result {
+            switch result {
+            case .failure:
                 completion(.failure(Error.connectivity))
+            case .success:
+                completion(.failure(Error.invalidData))
             }
         }
     }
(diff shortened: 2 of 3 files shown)


FeedAPIChallenge/Feed Feature/FeedImage.swift
```diff
@@ -17,3 +17,12 @@ public struct FeedImage: Hashable {
...
 }
+
+extension FeedImage: Decodable {
+    private enum CodingKeys: String, CodingKey {
+        case id = "image_id"
+        case description = "image_desc"
+        case location = "image_loc"
+        case url = "image_url"
+    }
+}
```
mapostolakis, line 28: The `CodingKeys` are API-specific details defined in the backend. So declaring the `CodingKeys` here will couple the `FeedImage` data model with API implementation details. And since other modules depend on the `FeedImage`, they'll also be coupled with API implementation details.

Ideally, the API keys should be defined in the Feed API module. So, for example, if there's a change in the backend, it doesn't propagate everywhere in the codebase. You just update the Feed API module without affecting others.

Suggestion: Create an API-specific struct in the Feed API module to perform the decoding. Thus, preventing API details from leaking into other modules.

mapostolakis (commented):
Hi @jeanm-dev,

I left one comment regarding low-level API-specific implementation details. Please, feel free to push a new commit.

Great job! ✅

FeedAPIChallenge/Feed Feature/FeedImage.swift
jeanm-dev: Hi Mike, 

I added the API-specific struct as per your suggestion. 
I noticed when I went back to the lectures that you introduced a separate mapper to deal with the mapping related logic.

Was the reasoning behind this to keep to the **single responsibility** principle?
Because it would be easy to add a private map function that just does the mapping inside the `RemoteFeedLoader`. 

For example:

_At the point of invocation_
```{swift}
switch result {
    case .failure:
        if response.statusCode == 200, let feedItems = try? JSONDecoder().decode(Root.self, from: data) {
            let feedImages: [FeedImage] = feedItems.items.map { Self.map(feedItem: $0) }
        } else {
        }
    }
```
_Example map function_
```{swift}
    private static func map(feedItem: RemoteFeedImage) -> FeedImage {
        return FeedImage(
            id: feedItem.id,
            description: feedItem.description,
            location: feedItem.location,
            url: feedItem.url)
    }
```

mapostolakis: Hi @jeanm-dev,

We did it for organizational purposes. The idea is to extract the mapping logic in a separate component, so it can be developed, maintained, reused, and so on. You'll see the benefits of this separation as you progress in the lectures.

But you can place the mapping code inside the `RemoteFeedLoader` as well. It's a choice ✅

PR #81, opened 2020-11-29
Saifur Rahman - Feed API Challenge

fromlucknow:
Hope I did allright this time.

### DIFF
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -23,10 +23,15 @@ public final class RemoteFeedLoader: FeedLoader {
 			switch result{
 			
 			case let .success((data, response)):
-				if let data = try? JSONSerialization.jsonObject(with: data, options: .allowFragments){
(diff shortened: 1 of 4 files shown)


caiozullo (commented):
Hi @fromlucknow,

You almost passed the challenge!

The implementation looks great, but there are warnings in the project.

Please, fix the warnings and push it again. Thank you.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,17 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 	
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		
+		client.get(from: url) { [weak self] (result) in
+			if self == nil{
+				return
+			}
```
oliver-perez, line 25: **suggestion:** I think you can use `guard self != nil else { return }` instead, since `guard` is intended to make the happy path easier to read by "guarding" you from the unwanted conditions. 🙂

PR #94, opened 2020-12-30
Wilmer Barrios - Feed API Challenge

wilbarrios:
I've added a `FeedImageMapper`, and even I didn't use `self` con result completion (on `load` function), a `[weak self]` was required to pass the final test.

### DIFF
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -19,6 +19,6 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 	
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		
+		client.get(from: url) { _ in }
 	}
 }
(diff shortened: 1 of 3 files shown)


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,44 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 	
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
+		client.get(from: url) { [weak self] result in
+			guard let _ = self else { return } // Even if I didn't use `self`, The use of `FeedImageMapper` did a completion handler...
```
caiozullo, line 23: You can simplify the guard without the `let`:

```
```

Also, you can remove the comment 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,44 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			switch result {
+			case .success(let (data, response)):
+				do {
+					let items = try FeedImageMapper.map(data, response)
+					completion(.success(items))
+				} catch {
+					completion(.failure(Error.invalidData))
+				}
+				break
```
caiozullo, line 32: The `break` is unnecessary in this context. You can remove it.

caiozullo (commented):
Hi @wilbarrios,

Great job! ✅

PR #101, opened 2021-01-07
Bogdan Poplauschi - Feed API Challenge

bpoplauschi:
# Intent
Implement loading a collection of images using TDD

# Implementation

- implemented the `RemoteFeedLoader` `load` async method
- by applying the Single Responsibility Principle, here are the resulted types and functions:
- internal helper class `FeedImagesMapper` takes care of mapping tuples `(Data, HTTPURLResponse)` to `FeedLoader.Result`
  - private nested structs `Root` and `Image` are TDOs (transport domain objects) used only internally by `FeedImagesMapper ` to parse the `Data` received
  - using those types instead of adding `Decodable` conformance on `FeedImage` so we don't spill the API details into higher level abstractions (`FeedImage`)
  - private helper generic function `jsonDecode` handles just the mapping of `Data` to `Decodable` instance(s)
- private `HTTPURLResponse` extension with `isValid() -> Bool` function makes the code more readable
- added an extra test `test_load_requestsDataFromURL` to assure `RemoteFeedLoader.url` property is used when loading (by using a different url than `test_loadTwice_requestsDataFromURLTwice`)

# Class Diagram

classDiagram

Note: the diagram was created using [mermaid](https://mermaid-js.github.io/mermaid/) which uses a markdown syntax. The `classDiagram.md` contains the "source code" which is usually easy to maintain if added to the workspace.
It's not as nice as one created using [draw.io](draw.io), but it's easier to maintain IMO.


FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,55 @@
...
+		guard response.isValid(),
+			  let root: Root = jsonDecode(from: data) else {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
+		return .success(root.feedImages)
+	}
+	
+	private static func jsonDecode<T: Decodable>(from data: Data) -> T? {
+		return try? JSONDecoder().decode(T.self, from: data)
+	}
```
caiozullo, line 46: This function doesn't need to be generic since it's only used in a single context. It could return `Root` directly to simplify the implementation and usage.

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,55 @@
...
+		let description: String?
+		let location: String?
+		let url: URL
+		
+		enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
```
caiozullo, line 29: The properties' names could match the API spec, so it wouldn't require the `CodingKeys`. This way, the code can be shorter/less verbose.

caiozullo (commented):
Hi @bpoplauschi,

- All tests are passing, so your implementation complies with the specs ✅
- Bonus points for the extra test and the diagram ✅

Great job!

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
bogdan-poplauschi-coda: I agree, I decided to be extra verbose since I don't use `_` inside entity names, but I guess for a limited private class this is fine. I will make the change.

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
bogdan-poplauschi-coda: That's right, I know I should have kept the implementation to the minimal necessary, it's just in this case I can easily see other jsons decoded like that in the `RemoteFeedLoader`, so I thought that's useful. I will revert it to a non-generic.

bogdan-poplauschi-coda:
Thanks @caiozullo for looking into this so quickly, I imagine today was a busy day with the live episode and all.
I pushed the changes as discussed.

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
bpoplauschi: That's right, I know I should have kept the implementation to the minimal necessary, it's just in this case I can easily see other jsons decoded like that in the `RemoteFeedLoader`, so I thought that's useful. I will revert it to a non-generic.

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
bpoplauschi: I agree, I decided to be extra verbose since I don't use the `_` char inside entity names, but I guess for a limited private class this is fine. I will make the change.

bpoplauschi:
Thanks @caiozullo for looking into this so quickly, I imagine today was a busy day with the live episode and all.
I pushed the changes as discussed.

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,48 @@
...
+		let image_url: URL
+		
+		var feedImage: FeedImage {
+			FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
+		}
+	}
+	
+	internal static func map(_ data: Data, from response: HTTPURLResponse) -> FeedLoader.Result {
+		guard response.isValid(),
+			  let root: Root = jsonDecode(from: data) else {
```
caiozullo, line 31: Since `jsonDecode` now returns `Root`, you can remove the explicit `: Root` type definition ✅

The method name could also be more explicit, such as `decodeRoot(from: data)`

bpoplauschi: I renamed it to `decodeRoot(from:)`, it reads better now `let root = decodeRoot(from: data)` 👍

caiozullo:
Nice one!

PR #103, opened 2021-01-10
Ivan Ornes - Feed API Challenge

ivanornes:


FeedAPIChallenge/Feed API/RemoteImagesLoader.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+//  Created by Ivan Ornes on 10/1/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal struct RemoteImagesLoader {
+	
+	internal static func getImages(_ data: Data) throws -> [FeedImage] {
```
mapostolakis, line 13: Consider the name you chose for this struct. "Loading" usually signifies operations with side-effects such as loading from the network or a database. But `getImages` is a mapping function with no side-effects. 

It maps data to `[FeedImage]` so it would be better to first rename the struct accordingly avoiding the suffix "Loader", i.e. `FeedImageMapper`.

Then you can also rename the function `getImages` to simply `map` to denote its purpose.

FeedAPIChallenge/Feed API/RemoteImagesLoader.swift
```diff
@@ -0,0 +1,17 @@
...
+		guard let dictionary = try JSONSerialization.jsonObject(with: data, options: .allowFragments) as? [String: Any] else { return [] }
```
mapostolakis, line 14: Check the return statement of this guard clause.

You're returning an empty array if the data serialization fails. However, an empty `FeedImage` array is a valid feed representation.

It would be best to throw an error communicating to the clients of this function what went wrong, which in this case is "invalid data".

FeedAPIChallenge/Feed API/RemoteImagesLoader.swift
mapostolakis, line 13: The `throws` you have marked this function with is never utilized.

FeedAPIChallenge/Feed API/RemoteImagesLoader.swift
```diff
@@ -0,0 +1,17 @@
...
+		return (dictionary["items"] as? [[String: Any]] ?? []).compactMap(RemoteImage.init).map { $0.feedImage }
```
mapostolakis, line 15: `dictionary["items"] as? [[String: Any]]` is part of the data validation/extraction process. 

If you can't access the items as `[[String: Any]]` then the data should be considered invalid and you should throw an error.

Include this check in the guard statement.

mapostolakis, line 15: By using `compactMap` and discarding `nil` results you end up with the same problem you got by returning an empty array in the `guard` statement or the `dictionary["items"] as? [[String: Any]] ?? []`. Check my comments above for a more detailed explanation.

If the data for a JSON feed image is invalid you shouldn't ignore it. Instead, you should notify the clients of this function that the mapping process failed by throwing an appropriate error.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -4,21 +4,39 @@
...
-	public init(url: URL, client: HTTPClient) {
+	public init(url: URL, client: WeakHTTPClient?) {
 		self.url = url
 		self.client = client
 	}
 	
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		
+		client?.get(from: url) { [weak self] result in
+			guard self?.client != nil else { return }
```
mapostolakis, line 25: Check if `self` is `nil` instead of the client. I suggest you also review the `test_load_doesNotDeliverResultAfterSUTInstanceHasBeenDeallocated` test in the lectures in case this isn't clear yet.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -4,21 +4,39 @@
...
+						let images = try RemoteImagesLoader.getImages(data)
+						completion(.success(images))
+					} else {
+						completion(.failure(Error.invalidData))
+					}
+				case .failure(_): completion(.failure(Error.connectivity))
+				}
+			} catch {
+				completion(.failure(Error.invalidData))
+			}
```
mapostolakis, line 39: Since it's the `RemoteImagesLoader.getImages` that throws you don't have to wrap the whole `load` body in a `do/catch` block. 

You can wrap only the part of the throwing function, reducing the indentation and making the code more manageable and easier to read.

For example:

```
switch result {
case .success((let data, let response)):
  if response.statusCode == 200 {
    do {
      let images = try RemoteImagesLoader.getImages(data)
      completion(.success(images))
    } catch {
    }
  } else {
  }
case .failure(_):
}

```

mapostolakis (commented):
Hi @ivanornes,

You almost passed the challenge!

Please, check my review comments and let me know when you're done so I can review it again.

Thank you!

ivanornes:
Hi Mike, 

I have just reviewed the code. 

Thanks for the great feedback!

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -6,19 +6,35 @@ import Foundation
 
 public final class RemoteFeedLoader: FeedLoader {
 	private let url: URL
-	private let client: HTTPClient
+	private var client: HTTPClient?
```
mapostolakis, line 9: The goal of the `RemoteFeedLoader` is to request data over HTTP and map it to a native feed. So this component can't function properly without an `HTTPClient`, thus it shouldn't be optional. 

You can revert to `let client: HTTPClient`.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -6,19 +6,35 @@ import Foundation
...
 	
 	public enum Error: Swift.Error {
 		case connectivity
 		case invalidData
 	}
 	
-	public init(url: URL, client: HTTPClient) {
+	public init(url: URL, client: HTTPClient?) {
```
mapostolakis, line 16: Please see my comment above and remove the optional.

mapostolakis (commented):
Hi @ivanornes, almost there!

Please, check my last couple of review comments and let me know once you're done so I can review it again. Thank you!

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,18 @@
...
+
+import Foundation
+
+internal struct FeedImageMapper {
+	
+	internal static func map(_ data: Data) throws -> [FeedImage] {
+		guard let dictionary = try JSONSerialization.jsonObject(with: data, options: .allowFragments) as? [String: Any],
+			  let imageItems = dictionary["items"] as? [[String: Any]] else { throw RemoteFeedLoader.Error.invalidData }
+		return try imageItems.map(RemoteImage.init).map { $0.feedImage }
+	}
```
mapostolakis, line 17: I suggest you also learn how to use `Codable` as a replacement to `JSONSerialization`. It will help you simplify your code and avoid multiple checks. You can centralize the mapping logic instead of spreading it in multiple places: e.g. this function and `RemoteImage.init`.

ivanornes:
Hi Mike, 

I have just made the two changes. I removed the HTTPClient optional and changed the JSONSerialization for JSONDecoder. 

Using the Codable Protocol was my first choice but it didn't worked the first time, so I used the JSONSerialization. Now I made it work with the Decodable protocol. 

Anyway, the code now looks much cleaner. 

Thanks for the feedback!

mapostolakis:
Hi @ivanornes,

Great job! ✅

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -6,7 +6,7 @@ import Foundation
 
 public final class RemoteFeedLoader: FeedLoader {
 	private let url: URL
-	private let client: HTTPClient
+	private var client: HTTPClient
```
mapostolakis, line 9: Revert to: 

```swift
```

ivanornes: Hi @mapostolakis,

Reverted the var to let. 

Thanks Mike! 😊

PR #111, opened 2021-01-23
Martyn Pękala - Feed API Challenge

martynpekala:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -8,6 +8,8 @@ public final class RemoteFeedLoader: FeedLoader {
 	private let url: URL
 	private let client: HTTPClient
 	
+	static let OK_200 = 200
```
mapostolakis, line 11: `static let OK_200 = 200` could be moved to the `FeedImageMapper` since it's only used there. Then it could also be private.

mapostolakis (commented):
Hi @martynpekala,

I left one comment about moving a constant to a private scope. Please, feel free to push a new commit.

Great job! ✅

PR #116, opened 2021-02-01
Francisco Conelli - Feed API Challenge

fconelli:
Just wanted to point out that I've created the FeedImageMapper type in a separate file. This is breaking the restrcition of only changing the RemoteFeedLoader.swift.


caiozullo:
Hi @fconelli,

Great job! ✅

PR #118, opened 2021-02-03
Anton Ilinykh - Feed API Challenge

ailinykh:


mapostolakis (commented):
Hi @ailinykh,

Great job! ✅

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,47 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		let image_desc: String?
+		let image_loc: String?
+		let image_url: URL
+		
+		var item: FeedImage {
+			return FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
+		}
+	}
+	
+	static var OK_200: Int { 200 }
```
mapostolakis, line 54: This property can be private since it's only used within the scope of `FeedMapper`.

ailinykh: Yeah, sure! Also class should be `final` as well

PR #122, opened 2021-02-04
Kubilay Erdogan - Feed API Challenge

kublaios:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,41 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				case 200:
+					if let images = FeedImageMapper.feedImages(from: data) {
+						completion(.success(images))
+					} else {
+						completion(.failure(Error.invalidData))
+					}
+				default:
+					completion(.failure(Error.invalidData))
+				}
+			}
```
mapostolakis, line 38: The status code check can be done in the `FeedImageMapper.feedImages`, so all the validation logic happens in a single component. 

This way, the code can be shorter and more concise, without the need to repeat `completion(.failure(Error.invalidData))` twice.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,41 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		}
+	}
+}
+
+private struct FeedImageMapper {
+	static func feedImages(from data: Data) -> [FeedImage]? {
+		guard let root = try? JSONSerialization.jsonObject(with: data) as? [String: [[String: String]]],
```
mapostolakis, line 45: I recommend using `Codable` instead of `JSONSerialization` as it leads to simpler and more expressible code without multiple optional checks.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,41 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			  let items = root["items"]
+		else { return nil }
+
+		return items.compactMap {
+			guard let uuIdString = $0["image_id"],
+				  let uuId = UUID(uuidString: uuIdString),
+				  let urlString = $0["image_url"],
+				  let url = URL(string: urlString)
+			else { return nil }
```
mapostolakis, line 54: It's important to notice that `compactMap` will discard any `nil` results. 

A `nil` result in this context means invalid data (an item with missing or invalid id/URL).

And if the mapping of any item fails for any reason, it should fail the whole operation because that'd be an invalid response (not following the API contract).

An expressive way of dealing with this failure would be to mark the `FeedImageMapper.feedImages` method with `throws` instead of returning optional `[FeedImage]?`.

If the mapping fails for any reason, it can throw `.invalidData`.

mapostolakis (commented):
Hi @kublaios, you almost passed the challenge!

kublaios:
Thanks, @mapostolakis! I pushed the changes after eliminating the shortcuts based on your feedback. Feel free to review again at your convenience.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,50 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			}
+		}
+	}
+}
+
+private struct FeedImageMapper {
+	private static let HTTP_OK = 200
+
+	static func result(from data: Data, response: HTTPURLResponse) -> FeedLoader.Result {
+		guard response.statusCode == Self.HTTP_OK
```
mapostolakis, line 39: `Self` can be omitted since the static `HTTP_OK` property is referenced in a static scope.

mapostolakis (commented):
Hi @kublaios,

I left a comment. Feel free to push a new commit.

Great job! ✅

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
kublaios: Thanks for the suggestion. This is purely based on my personal preference, I prefer not to omit `self`/`Self` because it feels more explicit to me. :)

PR #142, opened 2021-02-20
Jyothish Johnson- Feed API Challenge.

jyothishjohnson:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +21,48 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				
+				guard httpResponse.statusCode == 200 else{
+					self?.feedLoaderCompletion?(.failure(Error.invalidData))
+					return
+				}
+				
+				guard let responseData = try? JSONDecoder().decode(FeedImageItemsStruct.self, from: data) else {
+					self?.feedLoaderCompletion?(.failure(Error.invalidData))
+					return
+				}
```
mapostolakis, line 39: Since both guard statements complete with `.failure(Error.invalidData)` before returning, the expression can be combined into a single guard:

```suggestion
guard httpResponse.statusCode == 200,
	let responseData = try? JSONDecoder().decode(FeedImageItemsStruct.self, from: data) else {
	completion(.failure(Error.invalidData))
	return
}
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +21,48 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				
+				self?.feedLoaderCompletion?(.success(responseData.items.map{ $0.feedImage }))
+				
+			case .failure(let error):
+				
+				if (error as NSError).code == 0 && (error as NSError).domain == "Test" {
+					self?.feedLoaderCompletion?(.failure(Error.connectivity))
+				}
```
mapostolakis, line 47: In case of a client failure, the `RemoteFeedLoader` should *always* complete with `.failure(Error.connectivity)` regardless of the error type/code/domain/etc.

Another point to notice here is that the production code **should not** reference any code/values used in the tests. Production code should be agnostic of test code.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +21,48 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 	
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		
+		feedLoaderCompletion = completion
```
mapostolakis, line 24: Capturing the completion closure in a property allows it to be invoked outside the scope of the `load` method by any component that has access to it (notice how the `feedLoaderCompletion` is not `private`).

Instead, the completion argument can be invoked directly in the `load` method scope avoiding creating and managing extra state in the `RemoteFeedLoader` which can complicate the code.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +21,48 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		}
+	}
+}
+
+private struct FeedImageItemsStruct : Decodable {
+	let items : [FeedImageStruct]
+}
+
+private struct FeedImageStruct : Decodable {
+
```
mapostolakis, line 59: There is an extra line break here.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +21,48 @@ public final class RemoteFeedLoader: FeedLoader {
...
+					self?.feedLoaderCompletion?(.failure(Error.invalidData))
+					return
+				}
+				
+				guard let responseData = try? JSONDecoder().decode(FeedImageItemsStruct.self, from: data) else {
+					self?.feedLoaderCompletion?(.failure(Error.invalidData))
+					return
+				}
+				
+				self?.feedLoaderCompletion?(.success(responseData.items.map{ $0.feedImage }))
```
mapostolakis, line 41: There is a missing space between `map{` and the following brace.

mapostolakis (commented):
Hi @jyothishjohnson, you almost passed the challenge!

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
jyothishjohnson: Thanks, Mike, I will update likewise

jyothishjohnson:
Hi, Mike
I have updated the pull request with the suggested changes, please have a look

Thanks

mapostolakis:
Hi @jyothishjohnson,

Great job! ✅

PR #143, opened 2021-02-21
Saravanakumar - Feed API Challenge

sarawanakumar:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,34 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		var imageId: UUID
+		var imageDesc: String?
+		var imageLoc: String?
+		var imageUrl: URL
 		
+		func toFeedImage() -> FeedImage {
+			return FeedImage(id: imageId, description: imageDesc, location: imageLoc, url: imageUrl)
+		}
 	}
 }
```
mapostolakis, line 52: Since the `Root` type is only referenced in the `FeedItemsMapper` I suggest you move it to the same file. 

Moreover, by doing so, you'll be able to mark it as `private` since it's a mapping implementation detail and doesn't have to be visible by external components.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,34 @@ public final class RemoteFeedLoader: FeedLoader {
...
+	}
+}
+
+struct Root: Decodable {
+	var items: [Feed]
+	var feedItems: [FeedImage] {
+		return items.map { $0.toFeedImage() }
+	}
+	
+	struct Feed: Decodable {
```
mapostolakis, line 42: `Feed` is used to refer to a collection of items. 

A more suitable name for representing each item object natively would be `FeedItem`.

mapostolakis (commented):
Hi @sarawanakumar, you almost passed the challenge!

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
sarawanakumar: @mapostolakis Thanks for your valuable comments, they are really insightful. I have addressed them. Please have a look again. thanks.

mapostolakis:
Hi @sarawanakumar,

Great job! ✅

PR #160, opened 2021-03-07
Abdul Diallo - Feed API Challenge

adiallo22:


caiozullo:
Hi @adiallo22,

To pass the challenge, you need to fix some things:

1) You used the `xcode12_4` branch, but it seems like your local Xcode is an older version. 

For example, the `filePath` should work fine on `xcode12_4`. But it doesn't work on older versions.

So you shouldn't have modified the provided test helpers to use `file`. Instead, select the appropriate branch matching your local Xcode version and *re-do* the challenge.

---

2) The `FeedImage.swift` file shouldn't be changed. You should only change the `RemoteFeedLoader.swift` file to implement the load method.

---

Please **re-do** the challenge following all the instructions in the README file.

Let me know once you're done so I can review your solution again. Thank you 🙂

adiallo22:
Hi Ciao, thanks for reviewing my PR.
As mentioned, I updated my Xcode to 12.4 and only made changes on the `
RemoteFeedLoader.swift` file
Find below my new PR
https://github.com/essentialdevelopercom/ios-lead-essentials-feed-api-challenge/pull/161
Thank you

PR #166, opened 2021-03-26
Oleksii feed api challenge

lytvynovBogdanov:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,19 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			guard self != nil else { return }
+			
+			switch result {
+			case .failure:
+				completion(.failure(Error.connectivity))
+			case .success(let response):
+				guard response.1.statusCode == 200,
+					  let feedImages = try? JSONDecoder().decode(FeedImages.self, from: response.0) else {
+					return completion(.failure(Error.invalidData))
+				}
```
mapostolakis, line 32: This code can become more expressive by destructuring the `response` tuple:

```suggestion
			case let .success((data, response)):
				guard response.statusCode == 200,
					  let feedImages = try? JSONDecoder().decode(FeedImages.self, from: data) else {
					return completion(.failure(Error.invalidData))
				}
```
```suggestion
			case .success(let response):
				guard response.1.statusCode == 200,
					  let feedImages = try? JSONDecoder().decode(FeedImages.self, from: response.0) else {
					return completion(.failure(Error.invalidData))
				}
```

FeedAPIChallenge/Feed Feature/FeedImageDecoding.swift
```diff
@@ -0,0 +1,41 @@
...
+
+	private enum codingKeys: String, CodingKey {
+		case items
+	}
+	
+	public init(from decoder: Decoder) throws {
+		let container = try decoder.container(keyedBy: codingKeys.self)
+		
+		items = try container.decode([FeedImage].self, forKey: .items)
+	}
```
mapostolakis, line 22: The initializer is not required in this context. `Decodable` can automatically infer the initialization code resulting in less boilerplate.

FeedAPIChallenge/Feed Feature/FeedImageDecoding.swift
```diff
@@ -0,0 +1,41 @@
...
+//
+
+import Foundation
+
+struct FeedImages: Decodable {
+	let items: [FeedImage]
+
+	private enum codingKeys: String, CodingKey {
+		case items
+	}
```
mapostolakis, line 16: This enum is not required since the case `items` matches the `FeedImages.items` property name.

FeedAPIChallenge/Feed Feature/FeedImageDecoding.swift
```diff
@@ -0,0 +1,41 @@
...
+	
+	public init(from decoder: Decoder) throws {
+		let container = try decoder.container(keyedBy: codingKeys.self)
+		
+		id = try container.decode(UUID.self, forKey: .image_id)
+		description = try container.decodeIfPresent(String.self, forKey: .image_desc)
+		location = try container.decodeIfPresent(String.self, forKey: .image_loc)
+		url = try container.decode(URL.self, forKey: .image_url)
+	}
+}
```
mapostolakis, line 41: The `CodingKeys` are API-specific details defined in the backend. So declaring the `CodingKeys` here (in the Feed Feature) will couple the `FeedImage` data model with API implementation details. And since other modules depend on the `FeedImage`, they'll also be coupled with API implementation details.

Ideally, the API keys should be defined in the Feed API module. So, for example, if there's a change in the backend, it doesn't propagate everywhere in the codebase. You just update the Feed API module without affecting others.

Suggestion: Create an API-specific struct in the Feed API module to perform the decoding. Thus, preventing API details from leaking into other modules.

mapostolakis (commented):
Hi @lytvynovBogdanov, you almost passed the challenge!

1. Please, rename the PR to reflect the name shown in the README file instructions.
2. Check my comments and let me know once you're done so I can review it again. Thank you 🙂

lytvynovBogdanov:
Hi @mapostolakis, thank you for your review and feedback. I updated code based on your comments. Could you please have a look when you have time? 

Thanks

FeedAPIChallenge/Feed API/FeedImageDecoding.swift
```diff
@@ -0,0 +1,31 @@
...
+//
+
+import Foundation
+
+struct FeedImages: Decodable {
+	let items: [FeedImage]
+}
+
+extension FeedImage: Decodable {
+	private enum codingKeys: String, CodingKey {
```
mapostolakis, line 16: The `String` type is not required. You can remove.

mapostolakis (commented):
Hi @lytvynovBogdanov,

I left one comment and also I want to highlight the following:

I recommend creating your own mapping-specific struct with the `Codable` requirements for two reasons:

---

1. This way, the compiler can synthesize the implementation for you.

---

2. A type can only conform to a protocol once. So you may have issues if you try to compose two modules that define the same extension.


Here's a guideline to avoid this issue: 

"Don't make a type you don't own conform to a protocol you don't own."

In this case, the `FeedImage` struct belongs to the Feed module, and `Codable` belongs to the standard library module. So the API module doesn't own the `Codable` protocol or the `FeedImage` struct.

Otherwise, if another module also extends the `FeedImage` to implement the `Codable` protocol, you would get a compiler error when composing the modules in a project. 

That's why I recommend you create a mapping-specific model, such as an internal `CodableFeedImage` struct as shown in the lectures.

lytvynovBogdanov:
Thank you @mapostolakis for your comments. I didn't want to just copy past the solution from lecture and replace `FeedItem` to `FeedImage` :) 
I also like your solution much more, but wanted to hear comments from you about my old style. 
I've updated my code based on new style. Thank you for your review!

FeedAPIChallenge/Feed API/FeedImageDecoding.swift
```diff
@@ -0,0 +1,46 @@
...
+			FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
+		}
+	}
+	
+	private static var OK_200: Int { 200 }
+
+	internal static func map(_ data: Data, _ response: HTTPURLResponse) throws -> [FeedImage] {
+		guard response.statusCode == OK_200 else { throw RemoteFeedLoader.Error.invalidData }
+		return try JSONDecoder().decode(Root.self, from: data).items.map { $0.item }
+	}
```
mapostolakis, line 36: This method is not used. You can remove it.

mapostolakis (commented):
Nice one, @lytvynovBogdanov!

FeedAPIChallenge/Feed API/FeedImageDecoding.swift
lytvynovBogdanov: Ah, I missed it. Removed.

mapostolakis: Nice one!

PR #167, opened 2021-04-02
Carlos Marcelo Tonisso Junior - Feed API Challenge

CarlosTonisso:


FeedAPIChallenge/Feed API/Helper/FeedImageMapper.swift
```diff
@@ -0,0 +1,47 @@
...
+//
+
+import Foundation
+
+internal struct FeedImageMapper {
+	
+	private struct Root: Codable {
+		internal var items: [Item]
+		internal var images: [FeedImage] {
+			return items.compactMap({ $0.item })
```
mapostolakis, line 16: No need to `compactMap` since there are no optional values to discard. Use `map` instead.

```suggestion
			return items.map(\.item)
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,15 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 	
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		
+		client.get(from: url) { [weak self ]result in
```
mapostolakis, line 22: The spacing is broken here. Please fix it.

```suggestion
		client.get(from: url) { [weak self] result in
```

mapostolakis (commented):
Hi @CarlosTonisso,

Great job! ✅

mapostolakis:
Nice one!

PR #168, opened 2021-04-03
Olexiy Bezrodniy - Feed API Challenge

alexeybezrodniy:


FeedAPIChallenge/Feed API/FeedMapper.swift
```diff
@@ -0,0 +1,48 @@
...
+//  FeedMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Alix on 4/3/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+
```
mapostolakis, line 11: There is an extra line break here. You can remove it.

FeedAPIChallenge/Feed API/FeedMapper.swift
```diff
@@ -0,0 +1,48 @@
...
+internal class FeedMapper {
```
mapostolakis, line 12: The `FeedMapper` can be marked as `final` since it's not intended to be subclassed. At the moment, it only serves as a namespace to a static method.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,14 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 	
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		
+		client.get(from: url) { [weak self] (result) in
+			if self == nil { return }
+			switch result {
+			case .success((let data, let response)):
```
mapostolakis, line 25: You can destructure the values with a single `let` instead of two:

```suggestion
			case let .success((data, response)):
```

mapostolakis (commented):
Hi @alexeybezrodniy,

Great job! ✅

PR #171, opened 2021-04-08
Omran Khoja - Feed API Challenge

OmranK:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,40 @@ public final class RemoteFeedLoader: FeedLoader {
...
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
+			switch result {
+			case let .success((data, response)):
+				guard response.statusCode == 200, let items = try? JSONDecoder().decode(ImageMapper.ImageItems.self, from: data) else {
+					completion(.failure(Error.invalidData))
+					return
+				}
+				completion(.success(items.images))
```
mapostolakis, line 30: Since there is the `ImageMapper` helper for performing mappings you can create a function and move all the mapping logic to it as well:

```swift
private struct ImageMapper {

	...
	
	static func map(_ data: Data, response: HTTPURLResponse) -> FeedLoader.Result {
		      let items = try? JSONDecoder().decode(ImageItems.self, from: data) else {
		}

		return .success(items.images)
	}
}
```

Then the calling site would be:

```suggestion
				completion(ImageMapper.map(data, response: response))
```

mapostolakis (commented):
Hi @OmranK,

Great job! ✅

PR #173, opened 2021-04-10
Andreas Link - Feed API Challenge

LinkAndreas:


FeedAPIChallenge/Feed API/FeedImageEntity.swift
```diff
@@ -0,0 +1,25 @@
...
+}
+
+extension FeedImageEntity {
+	enum CodingKeys: String, CodingKey {
+		case id = "image_id"
+		case description = "image_desc"
+		case location = "image_loc"
+		case url = "image_url"
+	}
+}
```
mapostolakis, line 25: I suggest you move this model close to where it's being referenced, i.e. `Root` and `FeedImageModelMapper`, so all the mapping related code is declared together and you don't have to jump around files when you revisit your code.

Another option is to move the `Root` and `FeedImageModelMapper` components in this file and rename it to reflect the intent of the code, e.g. `FeedImageMapper.swift`.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +23,53 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+private struct Root {
+	let images: [FeedImageEntity]
+}
+
+extension Root: Decodable {
+	enum CodingKeys: String, CodingKey {
+		case images = "items"
 	}
 }
```
mapostolakis, line 75: There is no need for declaring the `CodingKeys` enum. 

Rename the `Root.images` property to `Root.items` and remove the extension.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +23,53 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+	private func handleSuccessfulResponseWithStatusCode200(data: Data) -> FeedLoader.Result {
+		do {
+			let root: Root = try JSONDecoder().decode(Root.self, from: data)
+			return .success(FeedImageModelMapper.map(from: root.images))
+		} catch {
+			return .failure(Error.invalidData)
+		}
+	}
+}
```
mapostolakis, line 50: This is mapping logic and you already have a component responsible for mapping, the `FeedImageModelMapper`.

You should move it to `FeedImageModelMapper.map` and remove this function from the `RemoteFeedLoader`.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +23,53 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard let self = self else { return }
+
+			switch result {
+			case let .success((data, response)) where response.statusCode == Constants.OK_200:
+				completion(self.handleSuccessfulResponseWithStatusCode200(data: data))
```
mapostolakis, line 31: Checking that the status code is correct is part of the mapping logic. You should move it to `FeedImageModelMapper.map`.

mapostolakis (commented):
Hi @LinkAndreas, you almost passed the challenge!

FeedAPIChallenge/Feed API/FeedImageEntity.swift
LinkAndreas: Hello @mapostolakis, I really appreciate your review and suggestions. I decided to go for your second option such that the `RemoteFeedLoader` does not need to deal with the `FeedImage`-related mapping and can leave this responsibility to the `FeedImageMapper`.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
LinkAndreas: Thanks a lot for bringing this up @mapostolakis. I moved the distributed mapping logic to the `FeedImageMapper`, such that the `RemoteFeedLoader` can have a single responsibility.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
LinkAndreas: Checking the status code is now part of the `FeedImageModelMapper`.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
LinkAndreas: My initial idea was to use a different naming than the one that is included in the API contract, i.e., `items`. But when thinking about it, it makes sense that the lower layer (API) does not need to deal with feature/domain-specific knowledge (like whether it is an image). Hence I dropped the `CodingKeys` and used `items` instead. Is my explanation correct @mapostolakis?

LinkAndreas:
Thanks a lot for your comments and suggestions, Mike. I have reworked my solution according to your feedback. Can you please have a look at it again  @mapostolakis?

FeedAPIChallenge/Feed API/FeedImageEntity.swift
mapostolakis: Nice one!

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
mapostolakis: That's it, @LinkAndreas. 

The `Root` and `Item` models describe the API structure.

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,53 @@
...
+		do {
+			let root: Root = try JSONDecoder().decode(Root.self, from: data)
+			let images: [FeedImage] = root.items.map { item in
+				return FeedImage(
+					id: item.id,
+					description: item.description,
+					location: item.location,
+					url: item.url
+				)
+			}
```
mapostolakis, line 46: The `map` function can be simplified.

First, you can move the creation of a `FeedImage` from an `Item` to a computed property in the `Item` struct:

```swift
	...
		
	var image: FeedImage {
		return FeedImage(
			id: id,
			description: description,
			location: location,
			url: url
		)
	}
}
```

Now you can map items to images as such:

```suggestion
			let images = root.items.map(\.image)
```

Also the type of `images` can be inferred so there is no need to explicitly specify it.

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,53 @@
...
+		let description: String?
+		let location: String?
+		let url: URL
+	}
+
+	static func map(data: Data, response: HTTPURLResponse) -> FeedLoader.Result {
+		guard response.statusCode == Constants.OK_200 else { return .failure(RemoteFeedLoader.Error.invalidData) }
+
+		do {
+			let root: Root = try JSONDecoder().decode(Root.self, from: data)
```
mapostolakis, line 38: The type of `root` can be inferred. There is no need to explicitly specify it.

```suggestion
			let root = try JSONDecoder().decode(Root.self, from: data)
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,53 @@
...
+					description: item.description,
+					location: item.location,
+					url: item.url
+				)
+			}
+
+			return .success(images)
+		} catch {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
```
mapostolakis, line 51: You can simplify the implementation of this function by grouping the branches returning `.failure(RemoteFeedLoader.Error.invalidData)`:

```suggestion
		guard response.statusCode == Constants.OK_200,
			  let root = try? JSONDecoder().decode(Root.self, from: data) else { return .failure(RemoteFeedLoader.Error.invalidData) }

		return .success(root.items.map(\.image))
```

mapostolakis (commented):
Hi @LinkAndreas,

Great job! ✅

PR #175, opened 2021-04-11
Lívia Vasconcelos - Feed API Challenge

liuvasconcelos:
Feed API Challenge needed to complete 'Networking Module' at 'iOS Lead Essentials' course.

### DIFF
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -19,6 +19,8 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 	
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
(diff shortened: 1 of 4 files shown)


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,15 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 	
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		
+		client.get(from: url) { [weak self] result in
+			guard let _ = self else { return }
```
mapostolakis, line 23: Since `self` is not used within the closure, `let` is not necessary.

It's simpler to check if it's no nil:

```suggestion
			guard self != nil else { return }
```

mapostolakis (commented):
Hi @liuvasconcelos,

Great job! ✅

PR #178, opened 2021-04-15
Adrian Bilescu - Feed API Challenge.

diti223:
Back on the track!

### DIFF
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -19,6 +19,13 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { result in
+			switch result {
+			case .failure:
+				completion(.failure(Error.connectivity))
+			default:
+				completion(.failure(Error.invalidData))
+			}
+		}
 	}
 }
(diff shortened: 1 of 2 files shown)


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,48 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		guard response.statusCode == 200 else {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
+
+		do {
+			let root = try JSONDecoder().decode(Root.self, from: data)
+			return .success(root.feedImages)
+		} catch {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
```
mapostolakis, line 64: You can simplify the implementation of this function by grouping the branches returning `.failure(RemoteFeedLoader.Error.invalidData)`:

```suggestion
		guard response.statusCode == 200, 
		    let root = try? JSONDecoder().decode(Root.self, from: data) else { return .failure(RemoteFeedLoader.Error.invalidData) }

 		return .success(root.feedImages)
```

mapostolakis (commented):
Hi @diti223,

Great job! ✅

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
diti223: That was one of my initial implementations but I had some Decoding errors and wanted to debug it and look into the thrown error.

PR #187, opened 2021-04-21
Vita Solomina - Feed API Challenge

Vitaa:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,56 @@ public final class RemoteFeedLoader: FeedLoader {
...
 	}
 
+	private static let successCode = 200
+
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
+			switch result {
+			case .success((let data, let response)):
```
caiozullo, line 27: The let can be defined outside the tuple so the expression can be shorter:

```suggestion
			case let .success((data, response)):
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,56 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				if response.statusCode == RemoteFeedLoader.successCode, let items = RemoteFeedImageMapper.map(data) {
+					completion(.success(items))
+				} else {
+					completion(.failure(Error.invalidData))
+				}
+			case .failure(_):
```
caiozullo, line 33: The `_` is not necessary.

It can be removed:

```suggestion
			case .failure:
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,56 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		case id = "image_id"
+		case description = "image_desc"
+		case location = "image_loc"
+		case url = "image_url"
+	}
+
+	let id: UUID
+	let description: String?
+	let location: String?
+	let url: URL
```
caiozullo, line 62: It's ok to use the same API property names in the model because it's modeling exactly the API response.

So the code can be shorter and clear. The CodingKeys is not needed - you can remove it:

```suggestion
	let image_id: UUID
	let image_desc: String?
	let image_loc: String?
	let image_url: URL
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,56 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+	func toFeedImage() -> FeedImage {
+		return FeedImage(id: id, description: description, location: location, url: url)
+	}
+}
+
+private extension Array where Element == RemoteFeedImage {
+	func toFeedImage() -> [FeedImage] {
```
caiozullo, line 70: This method could be named in plural since it converts the array into another array.

```suggestion
	func toFeedImages() -> [FeedImage] {
```

caiozullo (commented):
Hi @Vitaa,

Great job! ✅

Vitaa:
Thank you @caiozullo for your comments! Pushed the changes to address them

caiozullo:
Nice one!

PR #188, opened 2021-04-22
Paulo Silva - Feed API Challenge

paulobfsilva:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,22 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				do {
+					let items = try FeedImageMapper.map(data, response)
+					completion(.success(items))
+				} catch {
+					completion(.failure(Error.invalidData))
+				}
+			case .failure(Error.connectivity):
+				completion(.failure(Error.connectivity))
+			case .failure(_):
+				completion(.failure(Error.connectivity))
```
danillahtin, line 36: `HTTPClient` doesn't return `Error.connectivity`, this is `RemoteFeedLoader` implementation detail. Thus `.failure` will never contain `Error.connectivity`.

Also both failure cases do the same thing and can be combined.

Moreover `(_)` can be omitted. 

Remove one `.failure` case.

```suggestion
			case .failure:
				completion(.failure(Error.connectivity))
```

danillahtin (commented):
Hey, @paulobfsilva! 

Great job! ✅

PR #192, opened 2021-04-25
Prabhat Tiwari - Feed API Challenge

Prabhat07:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -5,6 +5,34 @@
 import Foundation
 
 public final class RemoteFeedLoader: FeedLoader {
+	
+	private struct Root: Codable {
```
mapostolakis, line 9: `Codable` is a type alias for the `Encodable` and `Decodable` protocols. In this case, we only need to decode `Root` from an external representation so `Codable` can be replaced with `Decodable`:

```suggestion
	private struct Root: Decodable {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -5,6 +5,34 @@
...
+		let items: [ImageItem]
+		var feed: [FeedImage] {
+			return items.map { $0.feedImage }
+		}
+	}
+
+	private struct ImageItem: Codable {
```
mapostolakis, line 16: Please see my comment about `Codable` above.

```suggestion
	private struct ImageItem: Decodable {
```

mapostolakis (commented):
Hi @Prabhat07, you almost passed the challenge!

PR #200, opened 2021-05-01
Fernando Vega - Feed API Challenge

fvegagiga:
Challenge finished!

### DIFF
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -19,6 +19,6 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { _ in }
 	}
 }
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -19,6 +19,8 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		client.get(from: url) { _ in }
+		client.get(from: url) { _ in
+			completion(.failure(Error.connectivity))
+		}
 	}
 }
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -21,8 +21,13 @@ public final class RemoteFeedLoader: FeedLoader {
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
 		client.get(from: url) { result in
 			switch result {
-			case .success:
-				completion(.failure(Error.invalidData))
+			case let .success((data, response)):
+				if response.statusCode == 200,
+				   let _ = try? JSONSerialization.jsonObject(with: data) {
+					completion(.success([]))
+				} else {
+					completion(.failure(Error.invalidData))
+				}
 			case .failure:
 				completion(.failure(Error.connectivity))
 			}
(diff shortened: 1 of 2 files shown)


mapostolakis:
Hi @fvegagiga, if your PR is ready for review please post a comment on the challenge page in the academy with the link to your PR. Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,54 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			return FeedImage(id: id, description: description, location: location, url: url)
+		}
+
+		enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
+	}
```
danillahtin, line 60: This is a struct that describes remote models. Style conventions for different languages are different so it is ok that remote model's properties don't follow swift's convention. Since this model comes not from swift but from other language and it is a private detail it is ok to violate swift's style convention and use underscore in the name. 
At a first glance it looks ok to create coding keys just to "fix" styling but when there is not one but 50-100 models it will feel different.

This is just a remark, no need to make the amendment.
```suggestion
	private struct Item: Decodable {
		let image_id: UUID
		let image_desc: String?
		let image_loc: String?
		let image_url: URL

		var item: FeedImage {
			return FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
		}
	}
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,54 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+	static var OK_200: Int { return 200 }
+
+	static func map(_ data: Data, _ response: HTTPURLResponse) throws -> [FeedImage] {
+		guard response.statusCode == OK_200 else {
+			throw RemoteFeedLoader.Error.invalidData
+		}
+
+		let root = try JSONDecoder().decode(Root.self, from: data)
+		return root.items.map { $0.item }
```
danillahtin, line 70: This method throws `invalidData` if `response` has invalid `statusCode` and it throws `DecodingError.dataCorrupted` when the data is invalid. It is a bit misleading that throwing `RemoteFeedLoader.Error.invalidData` means that `statusCode` is not `200` and doesn't meant that the data is actually invalid. 

Please either introduce a new descriptive error inside `FeedItemsMapper` to throw when status code is not `200` or throw `invalidData` on any mapping failure.

```suggestion
	static func map(_ data: Data, _ response: HTTPURLResponse) throws -> [FeedImage] {
		guard response.statusCode == OK_200, let root = try? JSONDecoder().decode(Root.self, from: data) else {
			throw RemoteFeedLoader.Error.invalidData
		}

		return root.items.map { $0.item }
```

danillahtin (commented):
Hi @fvegagiga, you almost passed the challenge!

Thank you 🙂

fvegagiga:
Hi @danillahtin ,
thank you for your review and feedback. I've updated my code based on your comments.

Best regards.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,47 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
+
+			switch result {
+			case let .success((data, response)):
+				if let items = try? FeedItemsMapper.map(data, response) {
+					completion(.success(items))
+				} else {
+					completion(.failure(Error.invalidData))
+				}
```
danillahtin, line 31: `FeedItemsMapper.map` is an internal detail of `RemoteFeedLoader` and we know that it can throw only `RemoteFeedLoader.Error.invalidData`. Current implementation is good. At the same time there is an implicit duplication of error handling. `try` here can be caught and thus error will be defined in one place that is `FeedItemsMapper.map`

Please check the suggestion.

```suggestion
				do {
					let items = try FeedItemsMapper.map(data, response)
					completion(.success(items))
				} catch {
					completion(.failure(error))
				}
```

danillahtin (commented):
Hi @fvegagiga,

Great job! ✅

PR #202, opened 2021-05-03
Chris Wong - Feed API Challenge.

therealchriswong:


FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,37 @@
...
+
+import Foundation
+
+internal final class FeedItemsMapper {
+	private struct Root: Decodable {
+		let items: [Item]
+		var feed: [FeedImage] {
+			return items.map { $0.item }
+		}
+	}
```
danillahtin, line 17: `internal` is a default access level. It is ok to explicitly declare it. On the other hand it is more important to be consistent across the codebase. Therefore if some class or method is explicitly marked as `internal` then all other should be marked as well. For example `Root.items` should be marked as well `Root.feed`. This project doesn't use explicit `internals` so it should be either added everywhere or removed here.

Please remove `internal` access modifier to follow the code convention of the codebase.
```suggestion
final class FeedItemsMapper {
	private struct Root: Decodable {
		let items: [Item]
		var feed: [FeedImage] {
			return items.map { $0.item }
		}
	}
```

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,37 @@
...
+		let image_desc: String?
+		let image_loc: String?
+		let image_url: URL
+
+		var item: FeedImage {
+			return FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
+		}
+	}
+
+	internal static func map(_ data: Data, from response: HTTPURLResponse) -> RemoteFeedLoader.Result {
```
danillahtin, line 30: Please remove internal access modifier to follow the code convention of the codebase.
```suggestion
	static func map(_ data: Data, from response: HTTPURLResponse) -> RemoteFeedLoader.Result {
```

danillahtin (commented):
Hi @therealchriswong,

I left a couple of review comments. Feel free to push changes.

Great job! ✅

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
therealchriswong: Good eye!

PR #204, opened 2021-05-07
Dennis Nehrenheim - Feed API Challenge

GrowthPilgrim:
- This PR implements the `RemoteFeedLoader` to load a collection of images from a backend.
- The `RemoteFeedLoader` conforms to the` <FeedLoader>` protocol creating an array of `FeedImage`.
- image
- The `RemoteFeedLoader` follows the backend contract (Remote Feed Image Spec) below:

| Property      | Type                |
|---------------|---------------------|
| `image_id`    | `UUID`              |
| `image_desc`  | `String` (optional) |
| `image_loc`   | `String` (optional) |
| `image_url`	| `URL`               |

### DIFF
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -19,6 +19,6 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { _ in }
 	}
 }
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -19,6 +19,8 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		client.get(from: url) { _ in }
+		client.get(from: url) { result in
+			completion(.failure(Error.connectivity))
+		}
 	}
 }
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -22,18 +22,18 @@ public final class RemoteFeedLoader: FeedLoader {
 		let items: [Image]
 
 		var feed: [FeedImage] {
-			return self.items.map { $0.item }
+			return self.items.map { $0.image }
 		}
 	}
 
 	private struct Image: Decodable {
-		let id: UUID
-		let description: String?
-		let location: String?
-		let image: URL
+		let image_id: UUID
+		let image_desc: String?
+		let image_loc: String?
+		let image_url: URL
 
-		var item: FeedImage {
-			return FeedImage(id: id, description: description, location: location, url: image)
+		var image: FeedImage {
+			return FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


mapostolakis:
Hi @GrowthPilgrim, if your PR is ready for review please post a comment in the challenge page in the academy with the link to your PR. Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,51 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard let self = self else { return }
```
danillahtin, line 23: No need to used `self` inside the block, thus no need to unwrap it. It is enough just to check if it is not nil.

Please replace conditional unwrapping with check for `nil`

```suggestion
			guard self != nil else { return }
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,51 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+			switch result {
+			case let .success((data, response)):
+				completion(FeedImagesMapper.map(data, from: response))
+
+			default:
```
danillahtin, line 29: `result` has only two cases: `.success` and `.failure`. Using `default` here can be confusing whereas using `.failure` is more explicit and strict. 

If in future `Result` type is required to have tree cases or more (`success`, `failure`, `otherCase`) then having only `.success` and `.failure` cases in switch will cause compilation error and will force the developer to fix it. On the other hand having `.success` and `default` will run smoothly and potentially can lead to a bug or to a mishandled case.

Please replace `default` case with `.failure`

```suggestion
			.failure:
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,51 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				completion(self.failure(.connectivity))
```
danillahtin, line 30: Private helper here only helps to omit `Error` type/namespace, on the other hand it adds 3 more lines of code and makes the developer to use `self`. Code would be much simpler if the helper is inlined.

Please inline the helper:

```suggestion
				completion(.failure(Error.connectivity))
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,51 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			}
+		}
+	}
+
+	private func failure(_ error: RemoteFeedLoader.Error) -> FeedLoader.Result {
+		return .failure(error)
+	}
```
danillahtin, line 37: The helper is not used, please remove it

```suggestion
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,51 @@ public final class RemoteFeedLoader: FeedLoader {
...
+}
+
+private final class FeedImagesMapper {
+	private struct Root: Decodable {
+		let items: [Image]
+
+		var feed: [FeedImage] {
+			return self.items.map { $0.image }
```
danillahtin, line 45: Please remove `self` to follow the codebase convention

```suggestion
			return items.map { $0.image }
```

danillahtin (commented):
Hi @GrowthPilgrim, you almost passed the challenge!

Thank you 🙂

GrowthPilgrim:
> Hi @GrowthPilgrim, you almost passed the challenge!
> 
> 
> 
> Thank you 🙂

Hey Danil, 

thx for the code review. I committed all your suggested changes locally and pushed them to the remote. I guess I also could have used this "Commit suggestion" feature of GitHub which I see for the first time. Would you suggest using that? I did not use it since there was no way for me to know if the suggestion builds and all tests pass. Or is the code change suggestion required to build and pass tests as well?

Best
Dennis

danillahtin (commented):
Hi @GrowthPilgrim,

Great job! ✅

danillahtin:
> I guess I also could have used this "Commit suggestion" feature of GitHub which I see for the first time. Would you suggest using that?

Commit suggestion would simply create a commit with suggestion. It will not check for compile time errors or rebuild and rerun the tests. 

If you are 100% sure that the change will not break anything then it is ok to commit the suggestion straight from the review page. 

On the other hand reimplementing the suggestion on your own will definitely help you to memorise and understand it better.

PR #206, opened 2021-05-12
Alex Lee - Feed API Challenge

hp0702:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -7,6 +7,9 @@ import Foundation
 public final class RemoteFeedLoader: FeedLoader {
 	private let url: URL
 	private let client: HTTPClient
+	private var OK_200: Int {
+		return 200
+	}
```
danillahtin, line 12: This code is not used, please remove it

```suggestion
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +22,60 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard let _ = self else {
+				return
+			}
```
danillahtin, line 28: `self` is not used inside this block, thus there's no need to unwrap it. Simple check if `self` is not `nil` is enough.

Please replace unwrapping with nullability check

```suggestion
			guard self != nil else {
				return
			}
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +22,60 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+			switch result {
+			case .success((let data, let response)):
```
danillahtin, line 31: The `let` can be defined outside the tuple so the expression can be shorter:

```suggestion
			case let .success((data, response)):
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +22,60 @@ public final class RemoteFeedLoader: FeedLoader {
...
+	}
+}
+
+final class FeedItemMapper {
+	private static var OK_200: Int {
+		return 200
+	}
+
+	static func map(_ data: Data, from response: HTTPURLResponse) -> FeedLoader.Result {
+		guard response.statusCode == self.OK_200 else {
```
danillahtin, line 46: Explicit `self` is not used in this codebase. Please remove `self` to make code consistent with the rest of the codebase.

```suggestion
		guard response.statusCode == OK_200 else {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +22,60 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+						let description = item["image_desc"] as? String
+						let location = item["image_loc"] as? String
+
+						result.append(FeedImage(id: uuid, description: description, location: location, url: url))
+					}
+
+					return .success(result)
+				}
+			}
```
danillahtin, line 74: It'd be best to use the `Swift.Decodable` APIs instead of the old `JSONSerialization` class. The code will be simpler and more concise.

Moreover, the goal of the challenges is to practice what you learn in the course. Please, replace `JSONSerialization` with `Decodable` as shown in the lectures to simplify the implementation.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +22,60 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			case .success((let data, let response)):
+				completion(FeedItemMapper.map(data, from: response))
+			case .failure:
+				completion(.failure(Error.connectivity))
+			}
+		}
+	}
+}
+
+final class FeedItemMapper {
```
danillahtin, line 40: This class is a `RemoteFeedLoader`'s implementation detail. No need for it to be accessible from other places than this file. 

Please make `FeedItemMapper` `private`.

```suggestion
private final class FeedItemMapper {
```

danillahtin (commented):
Hi @hp0702, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
hp0702: It makes sense to me. Updated.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
hp0702: Good to know. Updated.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
hp0702: Updated.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
hp0702: You're right. Updated.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
hp0702: Thanks for the suggestion. Updated

hp0702:
> Hi @hp0702, you almost passed the challenge!
> 
> 
> 
> Thank you 🙂

danillahtin (commented):
Hi @hp0702,

Great job! ✅

hp0702:
> Hi @hp0702,
> 
> Congratulations, you've passed the challenge 🥇
> 
> All tests are passing, so your implementation complies with the specs.
> 
> Great job! ✅

That's great. Thanks!!

PR #208, opened 2021-05-16
Manivel Nagarajan - Feed API Challenge

manivelnagarajan:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,44 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			case let .success((data, response)):
+				completion(FeedImageMapper.map(data: data, response: response))
+			case .failure:
+				completion(.failure(Error.connectivity))
+			}
+		}
+	}
+}
+
+class FeedImageMapper {
```
mapostolakis, line 34: The `FeedImageMapper` can be marked as `final` since it's not intended to be subclassed.

```suggestion
final class FeedImageMapper {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,44 @@ public final class RemoteFeedLoader: FeedLoader {
...
+	static let OK_200 = 200
```
mapostolakis, line 35: The `OK_200` property should have a private access level since it is used only in the `FeedImageMappper`.

```suggestion
	private static let OK_200 = 200
```

mapostolakis (commented):
Hi @manivelnagarajan,

Great job! ✅

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
manivel-nagarajan: Thanks, @mapostolakis for the feedback, I've made changes as per your suggestion.

manivelnagarajan: Thanks @mapostolakis for the feedback. I've pushed a new commit as per your suggestion.

mapostolakis:
Nice one, @manivelnagarajan 👍

PR #209, opened 2021-05-20
Saul Cortez - Feed API Challenge

SaulBeto:
Feed API Challenge Done :)

### DIFF
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -7,11 +7,14 @@ import Foundation
 public final class RemoteFeedLoader: FeedLoader {
 	private let url: URL
 	private let client: HTTPClient
-	private let code_200 = 200
 
 	public enum Error: Swift.Error {
 		case connectivity
 		case invalidData
+	}
+
+	enum Status: Int {
+		case code_200 = 200
 	}
 
 	public init(url: URL, client: HTTPClient) {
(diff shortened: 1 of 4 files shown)


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -13,12 +13,36 @@ public final class RemoteFeedLoader: FeedLoader {
 		case invalidData
 	}
 
+	enum Status: Int {
```
mapostolakis, line 16: The `Status` enum should have a private access level since it is used only internally in the private `map` function.

```suggestion
	private enum Status: Int {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -13,12 +13,36 @@ public final class RemoteFeedLoader: FeedLoader {
...
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+
+			guard let self = self else { return }
+
+			switch result {
+			case .success((let data, let response)):
```
mapostolakis, line 31: The `let` keyword can be defined once outside the tuple resulting in a shorter expression:

```suggestion
			case let .success((data, response)):
```

mapostolakis (commented):
Hi @SaulBeto,

Great job! ✅

PR #210, opened 2021-05-23
Karthik K Manoj - Feed API Challenge

karkoding:
Hi @caiozullo & @mapostolakis, here's my challenge!


FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,51 @@
...
+		private let description: String?
+		private let location: String?
+		private let image: URL
+
+		private enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case image = "image_url"
+		}
```
danillahtin, line 31: This is a struct that describes remote models. Style conventions for different languages are different so it is ok that remote model's properties don't follow swift's convention. Since this model comes not from swift but from other language and it is a private detail it is ok to violate swift's style convention and use underscore in the name.
At a first glance it looks ok to create coding keys just to "fix" styling but when there is not one but 50-100 models it will feel different.

This is just a remark, no need to make the amendment.

```suggestion
		private let image_id: UUID
		private let image_desc: String?
		private let image_loc: String?
		private let image_url: URL
```

danillahtin (commented):
Hi @karkoding,

Great job! ✅

FeedAPIChallenge/Feed API/FeedImageMapper.swift
karkoding: What an explanation. I believe I would not have got such an insight anywhere else. Just to make sure, for every remote model I can just name properties how it is defined in the remote correct? Hence there would not be any usage of `CodingKeys` enum?

I am making this change so that I learn and do not repeat this. Thank you.

danillahtin: Yes, it is ok to model responses with the same keys you get from remote as these models exist only to properly map the data from the remote to your local models. But it would not be a good thing if you start to use the models for something else, eg pass it to view models or to the database as it would create a dependency between the remote and other app components.

karkoding: Got it!
I have pushed my changes.
Thanks again.

danillahtin:
Nice one!

PR #217, opened 2021-05-27
Jose Alvarez - Feed API Challenge

daelhoser:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,46 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		let decoder = JSONDecoder()
+
+		if response.statusCode == 200, let images = try? decoder.decode(Items.self, from: data) {
+			return .success(images.feedImages)
+		} else {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
+	}
+
+	private struct Items: Codable {
```
mapostolakis, line 46: `Codable` is a typealias for the `Encodable` and `Decodable` protocols. In this case, we only need to decode types from an external representation, thus, we need only the `Decodable` conformance. 

You can replace `Codable` with `Decodable`:

```suggestion
	private struct Items: Decodable {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,46 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		private var items: [FeedItem]
+
+		var feedImages: [FeedImage] {
+			return items.map { $0.feedImage }
+		}
+	}
+
+	private struct FeedItem: Codable {
```
mapostolakis, line 54: `Codable` is a typealias for the `Encodable` and `Decodable` protocols. In this case, we only need to decode types from an external representation, thus, we need only the `Decodable` conformance. 

You can replace `Codable` with `Decodable`:

```suggestion
	private struct FeedItem: Decodable {
```

mapostolakis (commented):
Hi @daelhoser, you almost passed the challenge!

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,46 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			completion(self.map(response, withData: data))
+		}
+	}
+
+	// MARK: - Helper Methods
+
+	private func map(_ response: HTTPURLResponse, withData data: Data) -> FeedLoader.Result {
+		let decoder = JSONDecoder()
+
+		if response.statusCode == 200, let images = try? decoder.decode(Items.self, from: data) {
```
mapostolakis, line 39: `JSONDecoder` is only needed is the status code is 200. Otherwise, it'd be created even if never used.

As an optimization, create the `JSONDecoder` only if needed:

```suggestion
		if response.statusCode == 200, let images = try? JSONDecoder().decode(Items.self, from: data)
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,46 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			return .success(images.feedImages)
+		} else {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
+	}
+
+	private struct Items: Codable {
+		private var items: [FeedItem]
```
mapostolakis, line 47: `items` is never mutated so it can be a let:

```suggestion
		private let items: [FeedItem]
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,46 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		var feedImages: [FeedImage] {
+			return items.map { $0.feedImage }
+		}
+	}
+
+	private struct FeedItem: Codable {
+		private var image_id: UUID
+		private var image_desc: String?
+		private var image_loc: String?
+		private var image_url: URL
```
mapostolakis, line 58: The properties are never mutated so they can be declared as let instead of var:

```suggestion
		private let image_id: UUID
		private let image_desc: String?
		private let image_loc: String?
		private let image_url: URL
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
mapostolakis, line 47: `private` is not needed here since the struct is already private and other files can't access it. You can remove the private access level.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
mapostolakis, line 58: `private` is not needed here since the struct is already private and other files can't access it. You can remove the private access level.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
daelhoser: I went ahead and made the changes. Thank you

daelhoser:
@mapostolakis, all done. Please take another look. thanks.

mapostolakis:
Hi @daelhoser,

Great job! ✅

PR #218, opened 2021-05-28
Suhayl Ahmed - Feed API Challenge

Zanaeira:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,59 @@ public final class RemoteFeedLoader: FeedLoader {
...
+	private static func map(_ data: Data) -> [FeedImage]? {
+		var feedImages: [FeedImage] = []
+		do {
+			let items = try JSONDecoder().decode(Items.self, from: data)
+			feedImages = items.map()
+		} catch {
+			return nil
+		}
+
+		return feedImages
```
mapostolakis, line 50: The implementation of this function can be simplified by using a `try?` when decoding. If an error is thrown while evaluating the `try?` expression, the value of the expression will be `nil`:

```suggestion
		return try? JSONDecoder().decode(Items.self, from: data).map()
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,59 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		fileprivate let description: String?
+		fileprivate let location: String?
+		fileprivate let url: URL
+
+		enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
```
mapostolakis, line 74: The `FeedAPIImage` properties can be named exactly as the Feed API specs/contract. This way, the code is shorter and more concise since it doesn't need the `CodingKeys` enum.

mapostolakis (commented):
Hi @Zanaeira,

Great job! ✅

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,59 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			let items = try JSONDecoder().decode(Items.self, from: data)
+			feedImages = items.map()
+		} catch {
+			return nil
+		}
+
+		return feedImages
+	}
+
+	private final class Items: Decodable {
```
mapostolakis, line 53: This model type doesn't need to be a class because it doesn't need to be shared as a reference type. It can be a struct instead, since it only represents pure data:

```suggestion
	private struct Items: Decodable {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,59 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		fileprivate let items: [FeedAPIImage]
+
+		fileprivate func map() -> [FeedImage] {
+			return items.map {
+				FeedImage(id: $0.id, description: $0.description, location: $0.location, url: $0.url)
+			}
+		}
+	}
+
+	private final class FeedAPIImage: Decodable {
```
mapostolakis, line 63: This model type doesn't need to be a class because it doesn't need to be shared as a reference type. It can be a struct instead, since it only represents pure data:

```suggestion
	private struct FeedAPIImage: Decodable {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,59 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			feedImages = items.map()
+		} catch {
+			return nil
+		}
+
+		return feedImages
+	}
+
+	private final class Items: Decodable {
+		fileprivate let items: [FeedAPIImage]
```
mapostolakis, line 54: `fileprivate` is not needed here since the class is already private and other files can't access it. You can remove the fileprivate access level.

```suggestion
		let items: [FeedAPIImage]
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,59 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+		fileprivate func map() -> [FeedImage] {
```
mapostolakis, line 56: `fileprivate` is not needed here since the class is already private and other files can't access it. You can remove the fileprivate access level.

```suggestion
		func map() -> [FeedImage] {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,59 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				FeedImage(id: $0.id, description: $0.description, location: $0.location, url: $0.url)
+			}
+		}
+	}
+
+	private final class FeedAPIImage: Decodable {
+		fileprivate let id: UUID
+		fileprivate let description: String?
+		fileprivate let location: String?
+		fileprivate let url: URL
```
mapostolakis, line 67: `fileprivate` is not needed here since the class is already private and other files can't access it. You can remove the fileprivate access level.

```suggestion
		let id: UUID
		let description: String?
		let location: String?
		let url: URL
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
Zanaeira: Thanks so much for the feedback. It's all excellent but this point in particular was really useful. I really like the idea of going against Swift conventions just to show the actual intent. I would never have thought of that and it really does clean up the code nicely!

mapostolakis:
Nice one, @Zanaeira!

PR #222, opened 2021-05-30
Pol Piella Abadia - Feed API Challenge

pol-piella:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -4,6 +4,38 @@
...
+	var feedImage: FeedImage {
+		FeedImage(id: id, description: description, location: location, url: url)
+	}
+
+	enum CodingKeys: String, CodingKey {
+		case id = "image_id"
+		case description = "image_desc"
+		case location = "image_loc"
+		case url = "image_url"
+	}
```
mapostolakis, line 26: The `FeedItem` properties can be named exactly as the Feed API specs/contract. This way, the code is shorter and more concise since it doesn't need the `CodingKeys` enum.

mapostolakis (commented):
Hi @pol-piella,

Great job! ✅

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
pol-piella: Great shout! Pushed the change now 👍

mapostolakis: Nice one!

PR #223, opened 2021-05-31
Fei Wen - Feed API Challenge

fnytrip:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,47 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			case .failure:
+				completion(.failure(Error.connectivity))
+			case .success((let data, let response)):
+				completion(FeedImageMapper.map(from: data, response))
+			}
+		})
+	}
+	
+	private class FeedImageMapper {
+		static let OK_200: Int = 200
```
mapostolakis, line 34: The `OK_200` property should have a private access level since it is used only in the `FeedImageMappper`.

```suggestion
		private static let OK_200: Int = 200
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,47 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				return .failure(Error.invalidData)
+			} else {
+				let decoder = JSONDecoder()
+				do {
+					let root = try decoder.decode(Root.self, from: data)
+					return .success(root.items.map { $0.feedImage })
+				} catch  {
+					return .failure(Error.invalidData)
+				}
+			}
```
mapostolakis, line 47: You can simplify the implementation of this function by grouping the branches returning `.failure(Error.invalidData)`:

```suggestion
			guard response.statusCode == OK_200, let root = try? JSONDecoder().decode(Root.self, from: data) else {
				return .failure(Error.invalidData)
			}
			
			return .success(root.items.map { $0.feedImage })
```

mapostolakis (commented):
Hi @fnytrip,

I left only one a couple of comments. Feel free to push new changes.

Great job! ✅

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
fnytrip: Awesome, thanks, updated.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
fnytrip: Updated

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
mapostolakis: Nice one!

PR #224, opened 2021-05-31
Vlastimir Radojevic - Feed API Challenge

stovla:


FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,45 @@
...
+		var item: FeedImage {
+			FeedImage(id: id, description: description, location: location, url: url)
+		}
+
+		enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
```
mapostolakis, line 35: The `FeedItem` properties can be named exactly as the Feed API specs/contract. This way, the code is shorter and more concise since it doesn't need the `CodingKeys` enum.

mapostolakis (commented):
Hi @stovla,

Great job! ✅

PR #225, opened 2021-05-31
Prashant Korat - Feed API Challenge

koratprashant:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +21,52 @@ public final class RemoteFeedLoader: FeedLoader {
...
+					completion(.failure(Error.invalidData))
+				}
+			case .failure:
+				completion(.failure(Error.connectivity))
+			}
+		}
+	}
+}
+
+private class FeedImageMapper: Decodable {
```
mapostolakis, line 42: The `FeedImageMapper` doesn't need to conform to `Decodable` because it's not being decoded from an external representation.

```suggestion
private class FeedImageMapper {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +21,52 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		var item: FeedImage {
+			return FeedImage(id: id, description: description, location: location, url: image)
+		}
+
+		enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case image = "image_url"
+		}
```
mapostolakis, line 62: The `Item` properties can be named exactly as the Feed API specs/contract. This way, the code is shorter and more concise since it doesn't need the `CodingKeys` enum.

mapostolakis (commented):
Hi @koratprashant, you almost passed the challenge!

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
koratprashant: Thanks for your suggestions, added as per you mentioned.

koratprashant:
> Hi @koratprashant, you almost passed the challenge!
> 
> Please, check my comments and let me know once you're done so I can review it again. Thank you 🙂

Thanks for your feedback and time. I have push the changes.

mapostolakis:
Hi @koratprashant,

Great job! ✅

PR #226, opened 2021-05-31
Tan Tran - Feed API Challenge

tantan39:


FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,37 @@
...
+//  FeedAPIChallenge
+//
+//  Created by Tan Tan on 5/31/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+struct FeedImageMapper {
+	struct Root: Decodable {
```
caiozullo, line 12: The `Root` struct is only used inside the `FeedImageMapper`, so it can be private.

```suggestion
	private struct Root: Decodable {
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,37 @@
...
+		var items: [FeedImageResult] = []
+
+		struct FeedImageResult: Decodable {
```
caiozullo, line 15: The `FeedImageResult` struct is only used inside the `Root`, so it can be private.

```suggestion
		private struct FeedImageResult: Decodable {
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,37 @@
...
+			public let image_id: UUID
+			public let image_desc: String?
+			public let image_loc: String?
+			public let image_url: URL
```
caiozullo, line 19: Those properties are only used in the `FeedImageMapper` which has an `internal` access control. So making them public won't have any effect. You can remove the public keyword:

```suggestion
			let image_id: UUID
			let image_desc: String?
			let image_loc: String?
			let image_url: URL
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,14 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url, completion: { [weak self] result in
+			guard let _ = self else { return }
```
caiozullo, line 23: Since `self` is not used within the closure, `let` is not necessary.

It's simpler to check if it's no nil:

```suggestion
			guard self != nil else { return }
```

caiozullo (commented):
Hi @tantan39,

Great job! ✅

caiozullo:
Nice one!

PR #228, opened 2021-06-01
Aleksandr Honcharov - Feed API Challenge

Ariandr:


FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,48 @@
...
+//
+//  Created by Aleksandr Honcharov on 01.06.2021.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+enum FeedImagesMapper {
+	private struct Root: Decodable {
+		let items: [Image]
```
thiagopenna, line 13: The `items` property is not used outside of `Root`, so it can be private.
```suggestion
		private let items: [Image]
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,48 @@
...
+		var feed: [FeedImage] {
+			return items.map { $0.feedImage }
+		}
+	}
+
+	private struct Image: Decodable {
+		let id: UUID
+		let description: String?
+		let location: String?
+		let url: URL
```
thiagopenna, line 24: These properties are not used outside of `Image`, so they can be private.
```suggestion
		private let id: UUID
		private let description: String?
		private let location: String?
		private let url: URL
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,48 @@
...
+		var feedImage: FeedImage {
+			return FeedImage(id: id, description: description, location: location, url: url)
+		}
+
+		enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
```
thiagopenna, line 35: The `Image` properties can be named exactly as the Feed API specs/contract. This way, the code is shorter and more concise since it doesn't need the `CodingKeys` enum.

thiagopenna (commented):
Hi @Ariandr,

Great work! ✅

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
Ariandr: Thank you for the explanation, I'll add appropriate changes!

Ariandr:
Hi @thiagopenna 
Thank you for the review, I've committed all suggested changes.

thiagopenna:
Well done!

PR #229, opened 2021-06-02
Phillip Harris - Feed API Challenge

philliptharris:
I wasn't sure if it was "against the rules" to create additional files within the Feed API module, so I just kept all my code within RemoteFeedLoader.swift. Perhaps the code should be moved to a separate "mapper" class — but I think it is acceptable where it is.

### DIFF
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -19,6 +19,6 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { _ in }
 	}
 }
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -19,7 +19,8 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		client.get(from: url) { result in
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
 			completion(Swift.Result { try Self.mapClientResult(result) })
 		}
 	}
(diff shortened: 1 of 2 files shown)


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,47 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		let id: UUID
+		let description: String?
+		let location: String?
+		let url: URL
+		enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
```
danillahtin, line 59: The `RemoteImage` properties can be named exactly as the Feed API specs/contract. This way, the code is shorter and more concise since it doesn't need the `CodingKeys` enum.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,47 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			guard self != nil else { return }
+			completion(Swift.Result { try Self.mapClientResult(result) })
+		}
+	}
+
+	private static func mapClientResult(_ clientResult: HTTPClient.Result) throws -> [FeedImage] {
+		switch clientResult {
+		case .failure:
+			throw Error.connectivity
+		case .success((let data, let httpResponse)):
```
danillahtin, line 32: The `let` can be defined outside the tuple so the expression can be shorter:
```suggestion
		case let .success((data, httpResponse)):
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,47 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			guard httpResponse.statusCode == 200 else {
+				throw Error.invalidData
+			}
+			do {
+				let remoteItems = try JSONDecoder().decode(Items.self, from: data)
+				return remoteItems.items.map { $0.makeEquivalentFeedImage() }
+			} catch {
+				throw Error.invalidData
+			}
```
danillahtin, line 41: Since the error that is thrown by `JSONDecoder().decode` is ignored and simply replaced with `Error.invalidData`, `do { try JSONDecoder().decode } ... catch` block can be replaced with optional try like this:
```
guard let remoteItems = try? JSONDecoder().decode(Items.self, from: data) else {
    throw Error.invalidData
}
```

And if so, then two `guard ...` blocks can be combined into one to make the code shorter

```suggestion
			guard httpResponse.statusCode == 200, let remoteItems = try? JSONDecoder().decode(Items.self, from: data) else {
				throw Error.invalidData
			}
			return remoteItems.items.map { $0.makeEquivalentFeedImage() }
```

danillahtin (commented):
Hi @philliptharris,

Great job! ✅

PR #232, opened 2021-06-03
Lakshman - Feed API Challenge

Lakshman412:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,44 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			case let .success((data, response)):
+				completion(FeedImageMapper.map(data, response))
+			case .failure:
+				completion(.failure(Error.connectivity))
+			}
+		}
+	}
+}
+
+private final class FeedImageMapper {
```
danillahtin, line 35: `FeedImageMapper` instances has no specific behaviour. There's no value in initialising it but the code allows the developer to do so, `FeedImageMapper()` is absolutely valid code. 
In such cases when a type is used only as a namespace enums with no cases can be used to disallow instantiation at compile time so that `FeedImageMapper()` won't compile.

```suggestion
private enum FeedImageMapper {
```

danillahtin (commented):
Hi @Lakshman412,

Great job! ✅

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
Lakshman412: The suggested changes were made and pushed.

caiozullo: Nice one!

PR #233, opened 2021-06-03
Your Name - Feed API Challenge.

jainMadhur42:
I have done with the changes as required to complete the challenge looking forward to your response and review comments.

### DIFF
--- a/FeedAPIChallenge/Feed API/ImageFeedMapper.swift
+++ b/FeedAPIChallenge/Feed API/ImageFeedMapper.swift
@@ -8,9 +8,12 @@
 
 import Foundation
 
-class ImageFeedMapper {
+final class ImageFeedMapper {
 	private struct Root: Decodable {
 		let items: [Item]
+		var imageFeed: [FeedImage] {
+			items.map { $0.item }
+		}
 	}
 
 	private struct Item: Decodable {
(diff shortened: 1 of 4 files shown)


FeedAPIChallenge/Feed API/ImageFeedMapper.swift
```diff
@@ -0,0 +1,40 @@
...
+//  ImageFeedMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Madhur Jain on 03/06/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+final class ImageFeedMapper {
```
thiagopenna, line 11: `ImageFeedMapper` is defined as a class, so it's possible to instantiate it as `ImageFeedMapper()`. However, all its methods are static - it has no instance methods or properties.

You can make the initializer private in such cases since the `ImageFeedMapper` is just a namespace for static methods, with no instance-specific behavior.

Both solutions will disallow instantiation at compile time so that `ImageFeedMapper()` won't compile. ✅

thiagopenna (commented):
Hi @jainMadhur42 ,

Great job! ✅

PR #240, opened 2021-06-13
GordonFeng - Feed API Challenge

DearGordon:


FeedAPIChallenge/Feed API/FeedItemMapper.swift
```diff
@@ -0,0 +1,45 @@
...
+		var feedImage: FeedImage {
+			return FeedImage(id: self.image_id,
+			                 description: self.image_desc,
+			                 location: self.image_loc,
+			                 url: self.image_url)
+		}
+	}
+
+	private static let ok_code = 200
+	public static func mapping(_ data: Data, from response: HTTPURLResponse) -> Result<[FeedImage], Error> {
```
thiagopenna, line 33: As per my previous comment, there's no need to mark `mapping()` as `public`.
```suggestion
	static func mapping(_ data: Data, from response: HTTPURLResponse) -> Result<[FeedImage], Error> {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,18 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		client.get(from: url) { [weak self] receivedResult in
+			guard self != nil else { return }
+
+			switch receivedResult {
+			case let .success((receivedData, receivedResponse)):
+
+				let result = FeedItemMapper.mapping(receivedData, from: receivedResponse)
+				completion(result)
+
+			case .failure(_):
```
thiagopenna, line 31: The error value is not used so it can be ignored completely, making the code more concise.
```suggestion
			case .failure:
```

FeedAPIChallenge/Feed API/FeedItemMapper.swift
```diff
@@ -0,0 +1,45 @@
...
+//  FeedAPIChallenge
+//
+//  Created by Gordon Feng on 13/6/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+public final class FeedItemMapper {
+	private struct Root: Codable {
```
thiagopenna, line 12: `Codable` is a type alias for the `Encodable` and `Decodable` protocols. In this case, we only need to decode `Root` from an external representation so `Codable` can be replaced with `Decodable`:
```suggestion
	private struct Root: Decodable {
```

FeedAPIChallenge/Feed API/FeedItemMapper.swift
```diff
@@ -0,0 +1,45 @@
...
+		let items: [Item]
+		var feedImages: [FeedImage] {
+			return items.map({ $0.feedImage })
+		}
+	}
+
+	private struct Item: Codable {
```
thiagopenna, line 19: `Codable` is a type alias for the `Encodable` and `Decodable` protocols. In this case, we only need to decode `Item` from an external representation so `Codable` can be replaced with `Decodable`:
```suggestion
	private struct Item: Decodable {
```

thiagopenna (commented):
Hi @DearGordon ,

Great job! ✅

PR #241, opened 2021-06-13
Daniel Gallego Peralta - Feed API Challenge

danielios-hub:


FeedAPIChallenge/Feed API/RemoteImage.swift
```diff
@@ -0,0 +1,23 @@
...
+	let description: String?
+	let location: String?
+	let url: URL
+
+	enum CodingKeys: String, CodingKey {
+		case id = "image_id"
+		case description = "image_desc"
+		case location = "image_loc"
+		case url = "image_url"
+	}
```
thiagopenna, line 22: The `RemoteImage` properties can be named exactly as the Feed API specs/contract. This way, the code is shorter and more concise since it doesn't need the `CodingKeys` enum.

```suggestion
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,16 @@ public final class RemoteFeedLoader: FeedLoader {
...
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			switch result {
+			case let .success((data, response)):
+				let result = FeedImageMapper.getResult(from: data, response: response)
+				if self != nil {
+					completion(result)
+				}
```
thiagopenna, line 28: By checking if `self` has been deallocated only in the `success` path there's a possible odd behavior: `completion` might be called for the `failure` case, violating the specification that `RemoteFeedLoader`'s `load` method should not deliver a result after the instance has been deallocated.
Also, the code might be performing an expensive operation (`JSONDecoder().decode`), just to be discarded afterwards. I suggest checking for deallocation even before the switch statement on `result`.
```suggestion
				completion(result)
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,16 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			case .failure(_):
```
thiagopenna, line 29: The error value is not used so it can be ignored completely, making the code more concise.
```suggestion
			case .failure:
```

FeedAPIChallenge/Feed API/RemoteImage.swift
```diff
@@ -0,0 +1,23 @@
...
+//  RemoteImage.swift
+//  FeedAPIChallenge
+//
+//  Created by Daniel Gallego Peralta on 13/6/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal struct RemoteImage: Decodable {
```
thiagopenna, line 11: The convention in this codebase is to omit the default access level of `internal`, please remove it.
```suggestion
struct RemoteImage: Decodable {
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,43 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Daniel Gallego Peralta on 13/6/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+struct FeedImageMapper {
```
thiagopenna, line 11: `FeedImageMapper` is defined as a struct, so we can instantiate it as `FeedImageMapper()`. However, all its methods are static - it has no instance methods or properties.

It may be confusing for other developers to be able to instantiate a value type but then not be able to do anything with it.

As you'll see in the following lectures, you can make the initializer private in such cases since the `FeedImageMapper` works as a "namespace" for static methods, with no instance-specific behavior.

It could also be a `final class` with a private initializer, or an `enum` with no cases, which can never be instantiated!

All these solutions will disallow instantiation at compile time so that `FeedImageMapper()` won't compile ✅
```suggestion
struct FeedImageMapper {
	private init() {}
```

thiagopenna (commented):
Hi @danielios-hub ,

Great job! ✅

PR #245, opened 2021-06-22
Bryan Hoke - Feed API Challenge

BryanHoke:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,14 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard let _ = self else { return }
```
danillahtin, line 23: Since `self` is not used within the closure, `let` is not necessary.

It's simpler to check if it's not nil:

```suggestion
			guard self != nil else { return }
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,46 @@
...
+			FeedImage(id: id, description: description, location: location, url: url)
+		}
+
+		enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
+	}
```
danillahtin, line 35: The `Item` properties can be named exactly as the Feed API specs/contract. This way, the code is shorter and more concise since it doesn't need the `CodingKeys` enum.

```suggestion
	private struct Item: Decodable {
		let image_id: UUID
		let image_desc: String?
		let image_loc: String?
		let image_url: URL

		var image: FeedImage {
			FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
		}
	}
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,46 @@
+//
+//  FeedImagesMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Bryan Hoke on 6/21/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+final class FeedImagesMapper {
```
danillahtin, line 10: `FeedImagesMapper` is defined as a class, so we can instantiate it as `FeedImagesMapper()`. However, all its methods are static - it has no instance methods or properties.

As you'll see in the following lectures, you can make the initialiser private in such cases since the `FeedImagesMapper` is just a namespace for static methods, with no instance-specific behaviour.

danillahtin (commented):
Hi @BryanHoke,

Great job! ✅

PR #247, opened 2021-06-25
Rafael Asencio - Feed API Challenge

rafaelasencioB:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,58 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		let items: [Item]
+	}
+
+	private struct Item: Hashable, Decodable {
+		let id: UUID
+		let description: String?
+		let location: String?
+		let url: URL
+
+		var item: FeedImage {
```
danillahtin, line 50: Since in this context `item` means an image item from remote it is better to call this property as something like `image` or `local` so that there's no confusion in calling `item.item`.

```suggestion
		var local: FeedImage {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,58 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			return FeedImage(id: id, description: description, location: location, url: url)
+		}
+
+		init(id: UUID, description: String?, location: String?, url: URL) {
+			self.id = id
+			self.description = description
+			self.location = location
+			self.url = url
+		}
```
danillahtin, line 59: This initialiser will be autogenerated by the compiler, no need to declare it explicitly

```suggestion
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,58 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		})
+	}
+}
+
+private class FeedImageMapper {
+	private struct Root: Decodable {
+		let items: [Item]
+	}
+
+	private struct Item: Hashable, Decodable {
```
danillahtin, line 44: There's no need for the `Item` to conform to `Hashable`, please remove the conformance

```suggestion
	private struct Item: Decodable {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,58 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		let id: UUID
+		let description: String?
+		let location: String?
+		let url: URL
```
danillahtin, line 48: The `Item` properties can be named exactly as the Feed API specs/contract. This way, the code is shorter and more concise since it doesn't need the `CodingKeys` enum.

```suggestion
		let image_id: UUID
		let image_desc: String?
		let image_loc: String?
		let image_url: URL
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,58 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			self.location = location
+			self.url = url
+		}
+
+		private enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
```
danillahtin, line 66: To follow up on the previous comment, `CodingKeys` can be removed

```suggestion
```

danillahtin (commented):
Hi @rafaelasencioB, you almost passed the challenge!

Thank you 🙂

rafaelasencioB:
Hi! @danillahtin, thank you very much for the feedback. I have followed all the indications you told me, 
take a look when you can.
Thank you!

danillahtin (commented):
Hi @rafaelasencioB,

Great job! ✅

PR #248, opened 2021-06-25
Avelino Rodrigues - Feed API Challenge

avelino-rodrigues:


FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,41 @@
...
+//  FeedImagesMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Avelino Rodrigues on 25/06/2021.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedImagesMapper {
```
thiagopenna, line 11: `FeedImagesMapper` is defined as a class, so we can instantiate it as `FeedImagesMapper()`. However, all its methods are static - it has no instance methods or properties.

thiagopenna: Also, feel free to remove the default access level keyword `internal` to follow the project convention.

thiagopenna (commented):
Hi @avelino-rodrigues ,

Great job! ✅

PR #250, opened 2021-06-30
Ankit Kumar - Feed API Challenge.

ankitkumar-ios:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,63 @@ public final class RemoteFeedLoader: FeedLoader {
...
+					return
+				}
+
+			case .failure(_):
+				completion(.failure(Error.connectivity))
+			}
+		}
+	}
+
+	private class RemoteFeedItem: Decodable {
```
thiagopenna, line 53: This model type doesn't need to be a `class` because it doesn't need to be shared as a reference type. It can be a `struct` instead, since it only represents pure data. That way, we can also use `struct`'s default initializer, instead of having to declare one.
```suggestion
	private struct RemoteFeedItem: Decodable {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,63 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		let items: [RemoteFeedImage]
+		init(items: [RemoteFeedImage]) {
+			self.items = items
+		}
+	}
+
+	private class RemoteFeedImage: Decodable {
```
thiagopenna, line 60: See my previous comment:
```suggestion
	private struct RemoteFeedImage: Decodable {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,63 @@ public final class RemoteFeedLoader: FeedLoader {
...
+					completion(.failure(Error.invalidData))
+					return
+				}
+
+				do {
+					let value = try JSONDecoder().decode(RemoteFeedItem.self, from: data)
+					var feedImageList: [FeedImage] = []
+					value.items.forEach { item in
+						feedImageList.append(FeedImage(id: item.id, description: item.description, location: item.location, url: item.url))
+					}
```
thiagopenna, line 39: Instead of using a `forEach` and appending each item, it's possible to use a `map` for this transformation to avoid dealing with a mutable array, and also to make the code more concise and easier to follow.
```suggestion
					let feedImageList = value.items.map { item in
						return FeedImage(id: item.image_id, description: item.image_desc, location: item.image_loc, url: item.image_url)
					}
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,63 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			switch result {
+			case let .success((data, httpURLResponse)):
+
+				if httpURLResponse.statusCode != self.OK_200 {
+					completion(.failure(Error.invalidData))
+					return
+				}
+
+				do {
+					let value = try JSONDecoder().decode(RemoteFeedItem.self, from: data)
```
thiagopenna, line 35: Since we are ignoring the error being `thrown` by `JSONDecoder().decode`, there's no need to use a do/catch block, we can replace it with optional `try?`.
```suggestion
				if let value = try? JSONDecoder().decode(RemoteFeedItem.self, from: data) {
```
We can complete with `.failure(Error.invalidData)` in the `else` clause, and we will be able to combine the `statusCode` check in this `if` statement.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,63 @@ public final class RemoteFeedLoader: FeedLoader {
 		self.client = client
 	}
 
+	private var OK_200: Int { return 200 }
```
thiagopenna, line 21: The `OK_200` instance property can be static instead, since there's no need for an instance of `RemoteFeedLoader` to retrieve a constant value. It would be more suitable to have an instance property if the value varies per instance.
```suggestion
	private static var OK_200: Int { return 200 }
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,63 @@ public final class RemoteFeedLoader: FeedLoader {
...
+						feedImageList.append(FeedImage(id: item.id, description: item.description, location: item.location, url: item.url))
+					}
+					completion(.success(feedImageList))
+
+				} catch {
+					completion(.failure(Error.invalidData))
+					return
+				}
+
+			case .failure(_):
```
thiagopenna, line 47: Since we are ignoring the `error` in the `failure` case, we can omit it completely:
```suggestion
			case .failure:
```

thiagopenna (commented):
Hi @ankitkumar-ios , you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,63 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			self.location = location
+			self.url = url
+		}
+
+		enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
```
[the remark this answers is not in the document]
ankitkumar-ios: Hi, 
If any JSON keys changed from the backend then we have to update those things at 2-3 places. But by CodingKeys only one case needs to be updated.

Also I have made updated the changes. Thanks

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +18,44 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+				if httpURLResponse.statusCode != RemoteFeedLoader.OK_200 {
+					completion(.failure(Error.invalidData))
+					return
+				}
+
+				guard let value = try? JSONDecoder().decode(RemoteFeedItem.self, from: data) else {
+					completion(.failure(Error.invalidData))
+					return
+				}
```
thiagopenna, line 36: It's possible to combine both the `if` statement and the `guard` clause, having a single path for success and a single path for failure, instead of having two places that complete with `.invalidData.`
```suggestion
				guard httpURLResponse.statusCode == RemoteFeedLoader.OK_200,
				      let value = try? JSONDecoder().decode(RemoteFeedItem.self, from: data) else {
					return completion(.failure(Error.invalidData))
				}
```
Also, consider moving validation and/or mapping logic to helper functions, to make the `load(completion:)` function easier to read.

thiagopenna (commented):
Hi @ankitkumar-ios ,

The CI build failed.

Please, fix all build failures and let me know once you're done so I can continue the review. I also left an extra comment. Thank you 🙂

ankitkumar-ios:
Made the request changes. Please review further

caiozullo:
Hi @ankitkumar-ios, I think you forgot to push the changes. Please, push all changes so we can review them. Thank you!

ankitkumar-ios:
Sorry my bad, check for push was unchecked.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,44 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			return FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
+		}
+
+		static func map(_ data: Data, _ response: HTTPURLResponse) throws -> [FeedImage] {
+			guard response.statusCode == 200 else {
+				throw RemoteFeedLoader.Error.invalidData
+			}
+			let root = try JSONDecoder().decode(RemoteFeedItem.self, from: data)
+			return root.items.map({ $0.feedImageItem })
+		}
```
thiagopenna, line 60: This function is marked as `throwing`, but errors thrown here are ignored in the `load` function, I suggest two possible approaches:

---

`#1`: It's possible to make this function `throw` the correct type when the `JSONDecoder().decode` fails, that way we can use the error inside the `load` function (instead of ignoring it):
```swift
	static func map(_ data: Data, _ response: HTTPURLResponse) throws -> [FeedImage] {
		      let root = try? JSONDecoder().decode(RemoteFeedItem.self, from: data) else {
		}
		return root.items.map({ $0.feedImageItem })
	}
```

---

`#2`: It's possible to change the function return type to `FeedLoader.Result`, that way we don't need to use `throws` and can actually return an expected result, (which can  be used directly inside the `load` function):
```swift
	static func map(_ data: Data, _ response: HTTPURLResponse) -> FeedLoader.Result {
		      let root = try? JSONDecoder().decode(RemoteFeedItem.self, from: data) else {
		}
		return .success(root.items.map({ $0.feedImageItem }))
	}
````

File: FeedAPIChallenge/Feed API/RemoteFeedLoader.swift [hunk 10]
```diff
@@ -19,6 +19,44 @@ public final class RemoteFeedLoader: FeedLoader {
...
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
+			switch result {
+			case let .success((data, httpURLResponse)):
+				if let items = try? RemoteFeedImage.map(data, httpURLResponse) {
+					completion(.success(items))
+				} else {
+					completion(.failure(Error.invalidData))
+				}
```
thiagopenna, line 30: Following from my previous comment, here int eh `load` function we also have some possibilities:

---

`#1`: By making the function `throw` the correct error, we can use it here inside a `do/catch` block:
```swift
	do {
		let items = try RemoteFeedImage.map(data, httpURLResponse)
		completion(.success(items))
	} catch {
		completion(.failure(error))
	}
```
Or, we can use the `Result`'s `init(catching:)`, that receives a throwing closure as a parameter (perfect for this `map` function!)
```swift
	completion(Result { try RemoteFeedImage.map(data, httpURLResponse) })
```

---

`#2`: If on the other hand we decided to make the `map` function return a `Result`, we can use it directly:
```swift
	completion(RemoteFeedImage.map(data, httpURLResponse))
```

--- 

Also note how in both approaches we don't write `invalidData` in the `load` function: that's the `map`s function responsibility.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,44 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				completion(.failure(Error.connectivity))
+			}
+		}
+	}
+
+	private struct RemoteFeedItem: Decodable {
+		let items: [RemoteFeedImage]
+		init(items: [RemoteFeedImage]) {
+			self.items = items
+		}
```
thiagopenna, line 41: Since `RemoteFeedItem` is a `struct`, Swift provides a default initializer. There's no need to write a custom one that does the same thing.
```suggestion
```

thiagopenna (commented):
Hi @ankitkumar-ios ,

I left some extra review comments. Feel free to push changes.

Great job! ✅

ankitkumar-ios:
Thanks Penna, that was useful.
I have refactored the class to accommodate these changes

thiagopenna:
Well done!

PR #251, opened 2021-07-05
Josué Quiñones - Feed API Challenge

CobainJosue:


FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,45 @@
...
+		var images: [FeedImage] {
+			return items.map { $0.image }
+		}
+	}
+
+	private struct Item: Decodable {
+		public let imageId: UUID
+		public let imageDesc: String?
+		public let imageLoc: String?
+		public let imageURL: URL
```
danillahtin, line 24: `Item` is private so its properties should not be public, please remove `public` modifiers.

```suggestion
		let imageId: UUID
		let imageDesc: String?
		let imageLoc: String?
		let imageURL: URL
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,45 @@
...
+
+		enum CodingKeys: String, CodingKey {
+			case imageId = "image_id"
+			case imageDesc = "image_desc"
+			case imageLoc = "image_loc"
+			case imageURL = "image_url"
+		}
+
+		var image: FeedImage {
+			return FeedImage(id: imageId, description: imageDesc, location: imageLoc, url: imageURL)
```
danillahtin, line 34: `Item` is private so its properties should not be public, please remove `public` modifiers.

As shown in the lectures, the API data model properties can be named exactly as the Feed API specs/contract without the CodingKeys. This way, we can model the precise backend Feed API domain. 

Using underscores is not the convention in Swift, but in this case, this struct exists to describe remote API models with different conventions. So it is ok that the remote model's properties don't follow Swift's convention because it follows the remote model's convention. The code is more concise since it doesn't need the `CodingKeys` enum.

That's just a suggestion. If you and your team prefer using CodingKeys, that's fine too!

```suggestion
	private struct Item: Decodable {
		let image_id: UUID
		let image_desc: String?
		let image_loc: String?
		let image_url: URL

		var image: FeedImage {
			return FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
```

danillahtin (commented):
Hi @CobainJosue,

I left a couple of review comments. Feel free to push changes.

Great job! ✅

FeedAPIChallenge/Feed API/FeedImageMapper.swift
CobainJosue: Done, thank you!

danillahtin (commented):
Hi @CobainJosue!

Nice one!

PR #252, opened 2021-07-06
Vitaliy Lakhno - Feed API Challenge

LakhV:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +43,53 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+			case .failure(let error):
+				result = .failure(error)
+			}
+
+			completion(result)
+		}
+	}
+
+	private static func httpClientResult2RemoteFeedLoaderResult(_ httpClientResult: HTTPClient.Result) -> RemoteFeedLoaderResult {
```
thiagopenna, line 65: Please rename all methods to better follow the swift [guidelines](https://swift.org/documentation/api-design-guidelines/):

(I will suggest names for these functions, and will refer to them by their suggested names on follow-up comments)
```suggestion
	private static func remoteFeedLoaderResult(from httpClientResult: HTTPClient.Result) -> RemoteFeedLoaderResult {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +43,53 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			.flatMap { (responseData, httpResponse) -> RemoteFeedLoaderResult in
+				guard httpResponse.isStatusOK else {
+					return .failure(Error.invalidData)
+				}
+				return .success(responseData)
+			}
+		return feedLoaderResult
+	}
+
+	private static func responseData2FeedLoaderResult(_ responseData: Data) -> FeedLoader.Result {
```
thiagopenna, line 76: Following from my previous comment, suggested name:
```suggestion
	private static func feedLoaderResult(from responseData: Data) -> FeedLoader.Result {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -5,6 +5,30 @@
 import Foundation
 
 public final class RemoteFeedLoader: FeedLoader {
+	typealias RemoteFeedLoaderResult = Swift.Result<Data, Error>
```
thiagopenna, line 8: `RemoteFeedLoaderResult` is already inside the scope of `RemoteFeedLoader`, so there's no need to repeat the `RemoteFeedLoader` part in its name. 

Also, it's actually not the expected `Result` from `Remote`, it's a partial step into creating the expected `Swift.Result<[FeedImage], Error>`. A better name would describe this as the resulting `Data`, and we can also use the default `Swift.Error` to make it easier to map from `httpClientResult` (as shown in a later comment):
```suggestion
	typealias DataResult = Swift.Result<Data, Swift.Error>
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +43,53 @@ public final class RemoteFeedLoader: FeedLoader {
...
+	private static func httpClientResult2RemoteFeedLoaderResult(_ httpClientResult: HTTPClient.Result) -> RemoteFeedLoaderResult {
+		let feedLoaderResult = httpClientResult.mapError { _ in Error.connectivity }
+			.flatMap { (responseData, httpResponse) -> RemoteFeedLoaderResult in
+				guard httpResponse.isStatusOK else {
+					return .failure(Error.invalidData)
+				}
+				return .success(responseData)
+			}
+		return feedLoaderResult
+	}
```
thiagopenna, line 74: By changing the `RemoteFeedLoaderResult` to `DataResult`, we have to change it in this function. We can also change its name again to better represent the type returned. Also, there's no need to  create a variable for `feedLoaderResult`, we can return the result of `httpClientResult.mapError.flatMap` directly:
```suggestion
	private static func dataResult(from httpClientResult: HTTPClient.Result) -> DataResult {
		return httpClientResult.mapError { _ in Error.connectivity }
			.flatMap { (responseData, httpResponse) in
				guard httpResponse.isStatusOK else {
					return .failure(Error.invalidData)
				}
				return .success(responseData)
			}
	}
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +43,53 @@ public final class RemoteFeedLoader: FeedLoader {
...
+	private static func responseData2FeedLoaderResult(_ responseData: Data) -> FeedLoader.Result {
+		var result: FeedLoader.Result
+		do {
+			let remoteFeedImages = try JSONDecoder().decode(ResponseRootEntity.self, from: responseData)
+			let feedImages = remoteFeedImages.items.map { $0.feedImage }
+			result = .success(feedImages)
+		} catch {
+			result = .failure(Error.invalidData)
+		}
+		return result
```
thiagopenna, line 85: There's no need to use a `do/catch` block if we're ignoring the error `thrown` by `JSONDecoder().decode`, we can replace it with an optional `try?`. Also, there's no need to create a variable `result`, we can return any `Result` immediately:
```suggestion
		if let remoteFeedImages = try? JSONDecoder().decode(ResponseRootEntity.self, from: responseData) {
			let feedImages = remoteFeedImages.items.map { $0.feedImage }
			return .success(feedImages)
		}
		return .failure(Error.invalidData)
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +43,53 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			let feedLoaderResult = Self.httpClientResult2RemoteFeedLoaderResult(httpClientResult)
+			switch feedLoaderResult {
+			case .success(let responseData):
+				result = Self.responseData2FeedLoaderResult(responseData)
+
+			case .failure(let error):
+				result = .failure(error)
+			}
+
+			completion(result)
```
thiagopenna, line 61: Here, all the `switch` statement is doing is changing the success case from one type to another, while passing the same error. We can actually do that with `flatMap`:
```suggestion
			completion(Self.dataResult(from: httpClientResult)
				.flatMap(RemoteFeedLoader.feedLoaderResult))
```
(Also, note how `Self.dataResult(from: httpClientResult)` improves readability over `Self.httpClientResult2RemoteFeedLoaderResult(httpClientResult)`)

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +43,53 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			let feedImages = remoteFeedImages.items.map { $0.feedImage }
+			result = .success(feedImages)
+		} catch {
+			result = .failure(Error.invalidData)
+		}
+		return result
+	}
+}
+
+extension HTTPURLResponse {
```
thiagopenna, line 89: This extension is not supposed to be accessed from outside this file, so we should mark it as `private`
```suggestion
private extension HTTPURLResponse {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -5,6 +5,30 @@
...
+		var feedImage: FeedImage {
+			return .init(id: id, description: description, location: location, url: url)
+		}
+
+		enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
```
thiagopenna, line 29: As shown in the lectures, the API data model properties can be named exactly as the Feed API specs/contract without the CodingKeys. This way, we can model the precise backend Feed API domain. 

thiagopenna (commented):
Hi @LakhV, you almost passed the challenge!

Thank you 🙂

thiagopenna:
Hey @LakhV , do you need any help with your pull request?

LakhV:
@thiagopenna Fixed. Thanks for the feedback.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
LakhV: I prefer the way it's done, to work with well-formed model. Thanks for your comments.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +18,72 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+		let feedImages = remoteFeedImages.items.map { $0.feedImage }
+		return .success(feedImages)
+	}
+}
+
+//MARK: - HTTPURLResponse extension
+
+//NOTE: It would be better to share the extension below or at least move into separate file,
+//but due to restrictions specified in the task it's not allowed to add/change other files
```
thiagopenna, line 82: It could be better to share this extension, but on the other hand we cannot be sure that all endpoints across all use cases will follow the same logic for valid `statuscodes`. If in the future we have to deal with a different endpoint where `204` is valid, or `any 2xx`, this extension might get in the way.

Since this statusCode validation is an implementation detail of the `RemoteFeedLoader` (following the validation rules for the feed feature use case), it should not leak outside of this scope. Maybe it's something that we will solve in the future by having different modules per feature, but for the moment there's no need to share this extension.

thiagopenna (commented):
Hi @LakhV,

I left some suggestions and comments on different ways to improve this code. Feel free to push changes.

PR #253, opened 2021-07-06
Ahmad Atef - Feed API Challenge

ahmad-atef:


danillahtin:
Hi @ahmad-atef,

The CI build failed.

Please, fix all build failures and let me know once you're done so I can start the review. Thank you 🙂

ahmad-atef:
> Hi @ahmad-atef,
> 
> 
> The CI build failed.
> 
> Please, fix all build failures and let me know once you're done so I can start the review. Thank you 🙂

hey @danillahtin 
Thanks for the review, I guess it should be ready now 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -7,18 +7,28 @@ import Foundation
...
 		case connectivity
 		case invalidData
 	}
 
-	public init(url: URL, client: HTTPClient) {
+	public init(url: URL, client: HTTPClient, mapper: FeedImageMapper = RemoteFeedImageMapper.default) {
 		self.url = url
 		self.client = client
+		self.mapper = mapper
 	}
```
danillahtin, line 21: Since there will be no need to create an instance of `RemoteFeedImageMapper` to call a `static` method, we can remove it.
```suggestion
	public init(url: URL, client: HTTPClient) {
		self.url = url
		self.client = client
	}
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -7,18 +7,28 @@ import Foundation
...
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard let self = self else { return }
+			switch result {
+			case .success((let data, let response)):
+				completion(self.mapper.map(data, response: response))
```
danillahtin, line 28: Following from my previous comment, the `map` function could be `static` so there would be no need to create an instance of `RemoteFeedImageMapper ` to call it.

```suggestion
				completion(RemoteFeedImageMapper.map(data, response: response))
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -7,18 +7,28 @@ import Foundation
...
+			case .failure:
+				completion(.failure(RemoteFeedLoader.Error.connectivity))
```
danillahtin, line 30: `RemoteFeedLoader` can be omitted to make code shorter

```suggestion
				completion(.failure(Error.connectivity))
```

FeedAPIChallenge/Feed API/HTTPClient.swift
```diff
@@ -9,3 +9,27 @@ public protocol HTTPClient {
 
 	func get(from url: URL, completion: @escaping (Result) -> Void)
 }
+
```
danillahtin, line 12: It is not allowed by the instructions to change this file, please revert the change.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -7,18 +7,28 @@ import Foundation
...
+	public init(url: URL, client: HTTPClient, mapper: FeedImageMapper = RemoteFeedImageMapper.default) {
 		self.url = url
 		self.client = client
+		self.mapper = mapper
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard let self = self else { return }
```
danillahtin, line 25: Following from my next comment, there would be no reason to unwrap `self`, so it can be a simple nullability check
```suggestion
			guard self != nil else { return }
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -7,18 +7,28 @@ import Foundation
...
+			switch result {
+			case .success((let data, let response)):
```
danillahtin, line 27: The `let` can be defined outside the tuple so the expression can be shorter:

```suggestion
			case let .success((data, response)):
```

FeedAPIChallenge/Feed API/FeedMapper.swift
```diff
@@ -0,0 +1,53 @@
...
+//
+//  Created by Ahmed Atef Ali Ahmed on 02.07.21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+public protocol FeedImageMapper {
+	func map(_ data: Data, response: HTTPURLResponse) -> FeedLoader.Result
+}
```
danillahtin, line 13: Since `RemoteFeedLoader` doesn't need an instance conforming to `FeedImageMapper` to access `RemoteFeedImageMapper`'s static method, there's no need for this protocol. I suggest removing this protocol.

```suggestion
```

FeedAPIChallenge/Feed API/FeedMapper.swift
```diff
@@ -0,0 +1,53 @@
...
+
+struct Root: Decodable {
+	struct APIFeedImage: Decodable {
+		let image_id: UUID
+		let image_desc: String?
+		let image_loc: String?
+		let image_url: URL
+	}
+
+	var items: [APIFeedImage]?
```
danillahtin, line 38: `items` should not be optional according to the api spec, please make it non-optional.
Also there's no need for it to be mutable, please make it a `let` constant

```suggestion
	let items: [APIFeedImage]
```

FeedAPIChallenge/Feed API/FeedMapper.swift
```diff
@@ -0,0 +1,53 @@
...
+
+	public func map(_ data: Data, response: HTTPURLResponse) -> FeedLoader.Result {
+		guard response.statusCode == Self.OK_STATUS_CODE,
+		      let root = try? JSONDecoder().decode(Root.self, from: data)
+		else {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
+		return .success(root.images)
+	}
+}
```
danillahtin, line 28: `RemoteFeedImageMapper` is an internal implementation detail of `RemoteFeedLoader`, thus it should not be accessible from outside of the module, please make it `internal`

Following from my other comments, the `map` function could be made `static` so there would be no need to create an instance of `RemoteFeedImageMapper` to call it.

```suggestion
final class RemoteFeedImageMapper: FeedImageMapper {
	private static let OK_STATUS_CODE = 200

	static func map(_ data: Data, response: HTTPURLResponse) -> FeedLoader.Result {
		guard response.statusCode == OK_STATUS_CODE,
		      let root = try? JSONDecoder().decode(Root.self, from: data)
		else {
			return .failure(RemoteFeedLoader.Error.invalidData)
		}
		return .success(root.images)
	}
}
```

FeedAPIChallenge/Feed API/FeedMapper.swift
```diff
@@ -0,0 +1,53 @@
...
+
+struct Root: Decodable {
```
danillahtin, line 30: `Root` is an internal implementation detail of `RemoteFeedImageMapper` thus it should be private, please make it private

```suggestion
private struct Root: Decodable {
```

danillahtin (commented):
Hi @ahmad-atef, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/HTTPClient.swift
```diff
@@ -9,3 +9,27 @@ public protocol HTTPClient {
...
+			if let error = error {
+				completion(.failure(error))
+			} else if let data = data, let response = response as? HTTPURLResponse {
+				completion(.success((data, response)))
+			} else {
+				completion(.failure(RemoteHTTPClientError.noDataNoResponseNoError))
+			}
+		}
+	}
+}
```
danillahtin, line 35: It is not allowed by the instructions to change this file, please revert the change.
```suggestion
```

danillahtin (commented):
Hi @ahmad-atef, almost there!

Thank you 🙂

ahmad-atef:
> Hi @ahmad-atef, almost there!
> 
> 
> Thank you 🙂

Hi @danillahtin 
I've addressed your comments :)

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,46 @@
...
+//  FeedMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Ahmed Atef Ali Ahmed on 02.07.21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+final class FeedImageMapper {
```
danillahtin, line 11: `FeedImageMapper` is defined as a class, so we can instantiate it as `FeedImageMapper()`. However, all its methods are static - it has no instance methods or properties.

danillahtin (commented):
Hi @ahmad-atef,

Great job! ✅

PR #256, opened 2021-07-11
Yonic Surny - Feed API Challenge

aelptos:
Hi,

This is my PR for the Feed API Challenge.

Regards,
Yonic


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,14 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: self.url) { [weak self] result in
+			guard self != nil else { return }
+			switch result {
+			case let .success((data, response)):
+				completion(FeedImagesMapper.map(data, from: response))
+			default:
```
thiagopenna, line 27: We can improve readability by using an explicit `.failure` case, instead of `default`. Since we are dealing with a `Result` type and we don't expect it to change, we can better represent this logic by explicitly treating both the `.success` and `.failure`.
```suggestion
			case .failure:
```

FeedAPIChallenge/Feed Feature/FeedImagesMapper.swift
```diff
@@ -0,0 +1,48 @@
...
+//  FeedImagesMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Yonic Surny on 11/07/2021.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedImagesMapper {
```
thiagopenna, line 11: The convention in this codebase is to omit the default access level keyword `internal`. Please remove it.
```suggestion
final class FeedImagesMapper {
```

FeedAPIChallenge/Feed Feature/FeedImagesMapper.swift
```diff
@@ -0,0 +1,48 @@
...
+		}
+
+		var image: FeedImage {
+			return FeedImage(id: id, description: description, location: location, url: url)
+		}
+	}
+
+	private static var OK_200: Int { return 200 }
+
+	internal static func map(_ data: Data, from response: HTTPURLResponse) -> RemoteFeedLoader.Result {
```
thiagopenna, line 40: The convention in this codebase is to omit the default access level keyword `internal`. Please remove it.
```suggestion
	static func map(_ data: Data, from response: HTTPURLResponse) -> RemoteFeedLoader.Result {
```

FeedAPIChallenge.xcodeproj/project.pbxproj
```diff
@@ -123,8 +125,9 @@
 		08BBFF7C2492A5C100141A91 /* Feed Feature */ = {
 			isa = PBXGroup;
 			children = (
-				08BBFF7D2492A5C100141A91 /* FeedLoader.swift */,
 				08BBFF7F2492A5C100141A91 /* FeedImage.swift */,
+				0A51DA3A269AFC9B0077D77A /* FeedImagesMapper.swift */,
```
thiagopenna, line 129: `FeedImagesMapper` is an implementation detail of `RemoteFeedLoader`, so it should be grouped inside `Feed API` instead of `Feed Feature`.

thiagopenna (commented):
Hi @aelptos ,

Great job! ✅

FeedAPIChallenge/Feed Feature/FeedImagesMapper.swift
```diff
@@ -0,0 +1,48 @@
...
+		let description: String?
+		let location: String?
+		let url: URL
+
+		enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
```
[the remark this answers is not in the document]
aelptos: Hello @thiagopenna,

Sorry for the delay in the response, I was away on vacation.

Thank you for the detailed review and explanations.
I understand your point about breaking Swift naming conventions in this specific case.
In this exercice I would prefer keeping it with Coding Keys but this would certainly be discussed with the teams I'll work with.

Cheers,
Yonic

thiagopenna:
Well done!

PR #258, opened 2021-07-15
Angel Aguilar - Feed API Challenge.

AngelAguilar:


Tests/LoadFeedFromRemoteUseCaseTests.swift
```diff
@@ -24,108 +24,110 @@ class LoadFeedFromRemoteUseCaseTests: XCTestCase {
...
+
+		let items = [item1.model, item2.model]
+
+		expect(sut, toCompleteWith: .success(items), when: {
+			let json = makeItemsJSON([item1.json, item2.json])
+			client.complete(withStatusCode: 200, data: json)
+		})
+	}
+
 //
```
thiagopenna, line 117: Please remove extraneous commented lines on the `LoadFeedFromRemoteUseCaseTests` file.
```suggestion
		})
	}


```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,44 @@
...
+//  FeedImagesMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Angel Aguilar on 7/14/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedImagesMapper {
```
thiagopenna, line 11: `FeedImageMapper` is defined as a class, so we can instantiate it as `FeedImageMapper()`. However, all its methods are static - it has no instance methods or properties.

thiagopenna: The convention in this codebase is to omit the default access level keyword of `internal`, please remove it.
```suggestion
final class FeedImagesMapper {
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,44 @@
...
+				description: image_desc,
+				location: image_loc,
+				url: image_url
+			)
+		}
+	}
+
+	private static var OK_200: Int { 200 }
+
+	internal static func map(_ data: Data, from response: HTTPURLResponse) -> RemoteFeedLoader.Result {
```
thiagopenna, line 37: The convention in this codebase is to omit the default access level keyword of `internal`, please remove it.
```suggestion
	static func map(_ data: Data, from response: HTTPURLResponse) -> RemoteFeedLoader.Result {
```

thiagopenna (commented):
Hi @AngelAguilar ,

Great job! ✅

thiagopenna:
Well done!

PR #259, opened 2021-07-16
Ayaz Rahman - Feed API Challenge.

AyazRahman:


FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,49 @@
...
+		let description: String?
+		let location: String?
+		let imageURL: URL
+
+		enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case imageURL = "image_url"
+		}
```
thiagopenna, line 31: As shown in the lectures, the API data model properties can be named exactly as the Feed API specs/contract without the CodingKeys. This way, we can model the precise backend Feed API domain. 

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,49 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Ayaz Rahman on 16/7/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal struct FeedImageMapper {
```
thiagopenna, line 11: `FeedImageMapper` is defined as a struct, so we can instantiate it as `FeedImageMapper()`. However, all its methods are static - it has no instance methods or properties.

It may be confusing for other developers to be able to instantiate a struct but then not be able to do anything with it.

You can use a private initialize in a struct or a class, or even make it an enum with no cases, which can never be instantiated!

These solutions will disallow instantiation at compile time so that `FeedImageMapper()` won't compile ✅

thiagopenna: The convention in this codebase is to omit the default access leve keyword `internal`, please remove it.
```suggestion
struct FeedImageMapper {
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,49 @@
...
+		}
+
+		var feedImage: FeedImage {
+			return FeedImage(id: id, description: description, location: location, url: imageURL)
+		}
+	}
+
+	private static let OK_200: Int = 200
+
+	internal static func map(_ data: Data, from response: HTTPURLResponse) -> FeedLoader.Result {
```
thiagopenna, line 40: The convention in this codebase is to omit the default access leve keyword `internal`, please remove it.
```suggestion
	static func map(_ data: Data, from response: HTTPURLResponse) -> FeedLoader.Result {
```

thiagopenna (commented):
Hi @AyazRahman ,

Great job! ✅

thiagopenna:
Well done!

PR #260, opened 2021-07-23
Genrikh Beraylik - Feed API Challenge

beraylik:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +20,31 @@ public final class RemoteFeedLoader: FeedLoader {
...
+	private func handleSuccess(data: Data, response: HTTPURLResponse) -> FeedLoader.Result {
+		guard response.statusCode == OK_STATUS_CODE else {
+			return .failure(Error.invalidData)
+		}
+		do {
+			let items = try FeedLoaderMapper.map(data: data)
+			return .success(items)
+		} catch {
+			return .failure(Error.invalidData)
+		}
```
danillahtin, line 48: There's no need for explicit `do ... catch` block because the error that can emerge is not used. 
What this block does indeed is it maps the error that potentially can emerge on `try FeedLoaderMapper.map(data: data)` to `Error.invalidData`. It can be formulated in a more concise way using `try?`

Taking this into account and the `guard ... else` block at the top, the whole function can be shortened to the following:
```suggestion
		guard response.statusCode == OK_STATUS_CODE, let items = try? FeedLoaderMapper.map(data: data) else {
			return .failure(Error.invalidData)
		}
		return .success(items)
```

danillahtin (commented):
Hi @beraylik,

Great job! ✅

PR #261, opened 2021-07-26
Fernando Luiz Goulart - Feed API Challenge

fegoulart:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -13,12 +20,53 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			let items: [Image]
+		}
+
+		static func map(_ data: Data, from response: HTTPURLResponse) throws -> [Image] {
+			guard response.statusCode == OK_200, let root = try? JSONDecoder().decode(Root.self, from: data)
+			else {
+				throw RemoteFeedLoader.Error.invalidData
+			}
+			return root.items
+		}
```
danillahtin, line 64: This method can return not `[Image]` but `[FeedImage]` and do all the mapping inline using `Array.map`.
Thus `Image` can be placed inside `FeedImagesMapper` the same way as `Root` is placed and `Array` extension on lines `68`-`72` can be removed. 
The code would feel a bit cleaner as all the remote models are placed in the same namespace.

```suggestion
		static func map(_ data: Data, from response: HTTPURLResponse) throws -> [FeedImage] {
			guard response.statusCode == OK_200, let root = try? JSONDecoder().decode(Root.self, from: data)
			else {
				throw RemoteFeedLoader.Error.invalidData
			}
			return root.items.map { FeedImage(id: $0.image_id, description: $0.image_desc, location: $0.image_loc, url: $0.image_url) }
		}
```

danillahtin (commented):
Hi @fegoulart,

Great job! ✅

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
fegoulart: Perfect. It's done. Thank you very much !

danillahtin:
Hi @fegoulart, nice one!

PR #262, opened 2021-07-28
Muhammed Azharudheen- Feed API Challenge.

muh-azharudheen:


mapostolakis:
Hi @muh-azharudheen, if your PR is ready for review please post a comment on the challenge page in the academy with the link to your PR. Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,56 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			case let .success((data, response)):
+				guard let result = self?.feedLoaderResult(for: data, response: response) else { return }
+				completion(result)
+			}
+		}
+	}
+
+	// Helper
+	private func feedLoaderResult(for error: NSError) -> FeedLoader.Result? {
+		guard error == NSError(domain: "Test", code: 0) else { return nil }
```
thiagopenna, line 37: Your production logic should not be coupled with test code. The way this portion of code is currently implemented, if there's any error from the `HTTPClient`, the application will hang since the `load` method will not call `completion` with a `nil` `FeedLoader.Result`.

According to our tests, the loader `deliversConnectivityErrorOnClientError`, independently of which type of error.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,56 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		do {
+			let item = try JSONDecoder().decode(FeedImageResponse.self, from: data)
+			return .success(item.feedImages)
+		} catch {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
+	}
+
+	private func isInvalid(data: Data, response: HTTPURLResponse) -> Bool {
+		data == Data("invalid json".utf8) || response.statusCode != 200
```
thiagopenna, line 55: Following from my previous comment, by using `"invalid json"` this portion of code is leaking test code into production, and instead of testing against an artificial invalid data the function `isInvalid(data:response:)` should deal with all invalid cases.

The way it is currently implemented, this function does not verify if the data is indeed invalid, since the test `deliversInvalidDataErrorOn200HTTPResponseWithPartiallyValidJSONItems` will actually call the `JSONDecoder().decode` method.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,56 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] in
+			switch $0 {
```
thiagopenna, line 23: Instead of using shorthand argument names, we can improve readability bu using named arguments:
```suggestion
		client.get(from: url) { [weak self] httpResult in
			switch httpResult {
```

thiagopenna (commented):
Hi @muh-azharudheen, you almost passed the challenge!

Thank you 🙂

muh-azharudheen:
Hi @thiagopenna 
Resolved Comments. please check

Thanks you

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,46 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] httpResult in
+			switch httpResult {
+			case .failure:
+				completion(.failure(RemoteFeedLoader.Error.connectivity))
```
thiagopenna, line 25: `RemoteFeedLoader` can be omitted since this portion of code is in the same scope, making the code shorter:
```suggestion
				completion(.failure(Error.connectivity))
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,46 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			case let .success((data, response)):
+				guard let result = self?.feedLoaderResult(for: data, response: response) else { return }
```
thiagopenna, line 27: According to the specifications, the `RemoteFeedLoader` should not deliver any result after it has been deallocated. This implementation works for the client's `.success` case, but in the case of .`failure` it will violate the specs (since we're not checking for `nil` there).

I recommend moving the check of `self` being `nil` to before the `switch` statement. Something like:
```swift
guard let self = self else { return }
```
This way you can call `self.feedLoaderResult(for:response:)`

---
Or you could make the method `feedLoaderResult` `static` and simply check if `self` is `nil`:
```swift
```
That way you would call `RemoteFeedLoader.feedLoaderResult(for:response:)`

thiagopenna (commented):
Hi @muh-azharudheen , you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,46 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		guard response.statusCode == 200 else {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
+
+		do {
+			let item = try JSONDecoder().decode(FeedImageResponse.self, from: data)
+			return .success(item.feedImages)
+		} catch {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
```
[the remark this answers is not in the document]
muh-azharudheen: I believe that currently errors are not ignored as the error in decoding captures at catch block where we are returning invalid data. but since we have the same error for both decoding and non 200 status codes, we can simplify the code as above to make it simpler and more readable. committing as the above

muh-azharudheen:
@thiagopenna  resolved review comments. please check

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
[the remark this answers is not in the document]
thiagopenna: Sorry, I meant that the specific `error` instance `thrown` by `decode` is being ignored, not that we are ignoring the fact that an error occurred. We would not be ignoring the error if we used it directly or did some sort of pattern-matching against it, for instance:
```swift
	} catch {
		return .failure(error)
	}
```
In that case we would get specific errors of type `Swift.DecodingError.dataCorrupted` or `Swift.DecodingError.keyNotFound`, but we are _ignoring_ these instances and returning our own `Error.invalidData` inside the `.failure`.

thiagopenna (commented):
Hi @muh-azharudheen,

I answered one comment regarding the `JsonDecoder` errors, feel free to continue the discussion.

Great job! ✅

PR #264, opened 2021-08-03
Pavan Powani - Feed API Challenge

pavan288:


pavan288 (commented):
Self-review done

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,35 @@
...
+	var description: String?
+	var location: String?
+	var url: URL
+
+	enum CodingKeys: String, CodingKey {
+		case id = "image_id"
+		case description = "image_desc"
+		case location = "image_loc"
+		case url = "image_url"
+	}
```
thiagopenna, line 30: As shown in the lectures, the API data model properties can be named exactly as the Feed API specs/contract without the CodingKeys. This way, we can model the precise backend Feed API domain. 

thiagopenna (commented):
Hi @pavan288 ,

Great job! ✅

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,18 @@ public final class RemoteFeedLoader: FeedLoader {
...
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
+			switch result {
+			case let .success((data, response)):
+				if response.statusCode == 200, let images = try? JSONDecoder().decode(ImagesRoot.self, from: data) {
+					completion(.success(images.feedImages))
+				} else {
+					completion(.failure(Error.invalidData))
+				}
```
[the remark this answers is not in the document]
pavan288: Thanks for the suggestion Thiago!

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
pavan288: @thiagopenna 
Your point about code being written in different languages, and keeping it ubiquitous across the frontend and the backend is very valid, I hadn't thought of that, thank you!

I am still a little conflicted though. In my experience, backend APIs can sometimes return keys that are overly descriptive for their own benefits (this_is_an_overly_descriptive_response_key). And I like to transform that into variables that are more concise and just as descriptive as they should be without losing any context. Is that a bad idea or a bad sacrifice to make for consistency in the app code? Or should those decisions been made on a case-to-case basis?

In our case, the FeedImage was the model local to the app, and the 'Root' model was the one that mimics the backend and resides in the Mapper type. Eventually they both reside in the production code. So changes from the backend might require adding or removing properties from the main FeedImage type as well, which in turn propagates to all the parts of the app where the FeedImage type is referenced. So what is the actual benefit of maintaining two separate types (one being a transformer and the other being the actual type used to access data)?
Why not just use one type, that uses CodingKeys to do the transformation and its own variables to let other classes access its information?

caiozullo: Hi @pavan288, as Thiago said, that's just a suggestion. If you and your team prefer using CodingKeys, that's fine too!

But I highly recommend separating the API and app models at least.

I also recommend using the same API model names as the backend API returns to you because that API model represents exactly the model contract with the API.

And the FeedImage app model doesn't need to match the API model one to one. In fact, if something changes in the API model, nothing needs to change in the FeedImage app model. That's one of the benefits of this separation.

You just need to change the FeedImage app model if you want/need to add/remove new properties based on the backend. But that's an app change because of a new requirement, not because the backend has changed!

Another benefit is that we can completely replace the backend to point to another API with 100% different properties and semantic without changing anything on the app side. We just need to map the new API model into the local app model, and nothing else has to change. (A lot of developers think this is an unlikely scenario - but in my experience, this happened several times in mid to large projects for multiple reasons. The apps that were decoupled from API details were easy to migrate to new APIs. The others were practically impossible - requiring an almost complete rewrite of the app because it was too coupled with the old backend structure!).

So I recommend separating API infrastructure details (such as JSON keys) from app logic because they change for different reasons. Things that change for different reasons should be separate.

This is explained in detail in the lectures. I recommend you rewatch the Networking module ✅

pavan288: Thanks @caiozullo i'll do that! All in all, great stuff. Really enjoying the lectures :)

thiagopenna: Hey @pavan288, it's also worth considering that the backend refers to this endpoint as returning `FeedItem`s, so it could in the future represent other types apart from images.

If in the future we have to deal with a heterogeneous JSON with a property to define the object type, like `"type": "image"`, we won't need to change anything regarding our `FeedImage`, we would just change the remote representation and the mapping.

(We might need to change other parts of the app to support the new types, but as Caio said: "that's an app change because of a new requirement".)

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,42 @@
...
+//  FeedImagesMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Pavan Powani on 03/08/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+class FeedImagesMapper {
```
thiagopenna, line 11: `FeedImagesMapper` is not supposed to be subclassed, so it should be marked as `final`.
```suggestion
final class FeedImagesMapper {
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,42 @@
...
+	struct ImagesRoot: Decodable {
+		private var items: [Item]
+
+		var feedImages: [FeedImage] {
+			return items.map({ $0.feedImage })
+		}
+	}
+
+	fileprivate struct Item: Decodable {
```
thiagopenna, line 20: `Item` is not being accessed in any other part of this file, so it would be preferable to mark it as `private`.
```suggestion
	private struct Item: Decodable {
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,42 @@
...
+//  FeedAPIChallenge
+//
+//  Created by Pavan Powani on 03/08/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+class FeedImagesMapper {
+	struct ImagesRoot: Decodable {
```
thiagopenna, line 12: `ImagesRoot` is an implementation detail of `FeedImagesMapper`, so it should be marked as `private`.
```suggestion
	private struct ImagesRoot: Decodable {
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,42 @@
...
+			case url = "image_url"
+		}
+
+		var feedImage: FeedImage {
+			return FeedImage(id: id, description: description, location: location, url: url)
+		}
+	}
+
+	static func map(_ data: Data, from response: HTTPURLResponse) -> FeedLoader.Result {
+		guard response.statusCode == 200, let images = try? JSONDecoder().decode(ImagesRoot.self, from: data) else { return .failure(RemoteFeedLoader.Error.invalidData) }
```
thiagopenna, line 39: Suggestion: break this code into multiple lines to improve readability
```suggestion
		guard response.statusCode == 200, let images = try? JSONDecoder().decode(ImagesRoot.self, from: data) else { 
			return .failure(RemoteFeedLoader.Error.invalidData) 
		}
```

thiagopenna (commented):
Well done @pavan288!

PR #265, opened 2021-08-04
Raphael Haase - Feed API Challenge

rph8:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,54 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				completion(.failure(Error.connectivity))
+				return
+			}
+
+			guard let (data, httpResponse) = try? result.get(), httpResponse.statusCode == 200 else {
+				completion(.failure(Error.invalidData))
+				return
+			}
+
+			guard let jsonDict = try? JSONSerialization.jsonObject(with: data, options: []) as? NSDictionary, let itemsArr = jsonDict["items"] as? NSArray else {
```
thiagopenna, line 33: I recommend using `Codable` instead of `JSONSerialization` as it leads to simpler and more expressible code without multiple optional checks, as shown in the lectures.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,54 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				return
+			}
+
+			// Convert the JSON to objects
+			let feedImages = itemsArr.map { item -> FeedImage? in
+				guard let dict = item as? NSDictionary,
+				      let imageIdString = dict["image_id"] as? String,
+				      let imageId = UUID(uuidString: imageIdString),
+				      let urlString = dict["image_url"] as? String,
+				      let url = URL(string: urlString) else { return nil }
```
thiagopenna, line 57: A `nil` result in this context means invalid data (an item with invalid `id` or `URL`). In that case we should complete with a `.failure(Error.invalidData)`, but in this case the item will simply be ignored in the `reduce` logic below.

Please change this logic to account for invalid `id`s or `URL`s (and feel free to add a test for that 😄)

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,54 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				.map { item -> Bool in
+					guard let dict = item as? NSDictionary else { return false }
+					return dict["image_id"] != nil && dict["image_url"] != nil
+				}
+				.reduce(true) { previous, next in previous && next }
+
+			guard isValid else {
+				completion(.failure(Error.invalidData))
+				return
+			}
```
thiagopenna, line 49: Since this portion of logic is simply checking for `nil` instead of actually validating the values, you could use `JSONSerialization.isValidJSONObject(data)` instead. From the [documentation](
https://developer.apple.com/documentation/foundation/jsonserialization/1418461-isvalidjsonobject):

> _Returns a Boolean value that indicates whether the serializer can convert a given object to JSON data._

Or you could also leave the JSON validation to the actual conversion - so you don't need to traverse the dictionary twice.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,54 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				      let url = URL(string: urlString) else { return nil }
+				return FeedImage(id: imageId,
+				                 description: dict["image_desc"] as? String,
+				                 location: dict["image_loc"] as? String,
+				                 url: url)
+			}
+			.reduce([]) { previous, feedImage -> [FeedImage] in
+				guard let feedImage = feedImage else { return previous }
+				return previous + [feedImage]
+			}
```
thiagopenna, line 66: If we needed to remove the `nil` values from the array, Instead of using `reduce` we could use `compactMap`, as per the [documentation](https://developer.apple.com/documentation/swift/sequence/2950916-compactmap):

> _Returns an array containing the non-nil results of calling the given transformation with each element of this sequence._

```suggestion
			.compactMap { $0 }
```

Unfortunately that's not the solution in our case, since we should complete with `.failure(Error.invalidData)` if any item failed to map.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,54 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+			guard let _ = self else { return }
```
thiagopenna, line 68: Instead of using a `guard let` and ignoring the attribution, it's clearer to check for `nil`:
```suggestion
			guard self != nil else { return }
```
Also, this check should be performed before any completion calls, to avoid performing unnecessary work and to avoid completing with a failure after the `RemoteFeedLoader` has been deallocated.

thiagopenna (commented):
Hi @rph8 , you almost passed the challenge!

I really recommend that you try this challenge using `Codable` to compare how simpler the code becomes without multiple optional checks.

Also, check my review comments and let me know once you're done so I can review it again. 

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +30,34 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				let feedImages = response.items.map {
+					FeedImage(id: $0.imageId,
+					          description: $0.imageDesc,
+					          location: $0.imageLoc,
+					          url: $0.imageUrl)
+				}
+				completion(.success(feedImages))
+			} catch {
+				completion(.failure(Error.invalidData))
+			}
```
thiagopenna, line 60: Instead of using a `try` inside a `do/catch` block, since this portion of code does not care about any errors thrown by `JSONDecoder`, it's possible to achieve the same result with an optional `try?`.
```suggestion
			let decoder = JSONDecoder()
			decoder.keyDecodingStrategy = .convertFromSnakeCase
			guard let response = try? decoder.decode(FeedImageResponse.self, from: data) else {
				return completion(.failure(Error.invalidData))
			}
			
			let feedImages = response.items.map {
				FeedImage(id: $0.image_id,
						  description: $0.image_desc,
						  location: $0.image_loc,
						  url: $0.image_url)
			}
			completion(.success(feedImages))
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -5,6 +5,17 @@
 import Foundation
 
 public final class RemoteFeedLoader: FeedLoader {
+	struct FeedImageStruct: Codable {
+		var imageId: UUID
+		var imageUrl: URL
+		var imageDesc: String?
+		var imageLoc: String?
```
thiagopenna, line 12: As shown in the lectures, the API data model properties can be named exactly as the Feed API specs/contract. This way, we can model the precise backend Feed API domain. 

But that's just a suggestion. This way it will be easier to use the `JSONDecoder` since there will be no need to set `keyDecodingStrategy`, or use CodingKeys. But if you and your team prefer one way or the other, that's fine!
```suggestion
		var image_id: UUID
		var image_url: URL
		var image_desc: String?
		var image_loc: String?
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +30,34 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+			guard case .success = result else {
+				completion(.failure(Error.connectivity))
+				return
+			}
+
+			guard let (data, httpResponse) = try? result.get(), httpResponse.statusCode == 200 else {
+				completion(.failure(Error.invalidData))
+				return
+			}
```
thiagopenna, line 44: Instead of checking twice if `result` represents a `success` (with `case .success = result` and later with `try? result.get()`, it's possible to perform this operation once:
```suggestion
			guard let (data, httpResponse) = try? result.get() else {
				return completion(.failure(Error.connectivity))
			}

			guard httpResponse.statusCode == 200 else {
				return completion(.failure(Error.invalidData))
			}
```
Another possibility would be to use a `switch` statement on the result, and extract the map logic to a helper function, as shown in the program:
```swift
	switch result {
		completion(RemoteFeedLoader.map(data, from: response))
	case .failure:
	}
```
That way, we can separate `loading` logic (calling `client.get` and dealing with its `result`) from `mapping` logic (converting this result into a `FeedLoader.Result`)

thiagopenna (commented):
Hi @rph8 , you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
rph8: I don't think this is the best way to deal with this. Either you are using a backend language like JS that follows the same case conventions as Swift or if you use Python, that is why keyDecodingStrategy was introduced in Swift. I do not quite see the benefit of potentially keeping the exact naming convention of a backend written in Python vs. using keyDecodingStrategy.

thiagopenna: Hey @rph8, no problem at all, it was a suggestion. My reasoning behind it has two main aspects:

- First, imagine a scenario where in the future this endpoint gets a new property, and that property does not follow the snake case naming convention. It might seem like a far stretch, but the API is a volatile dependency outside of our control. So, if that happens, we will have to change existing code outside of `FeedImageStruct`, and that would mean that some other class has more than one reason to change (in violation of the Single Responsibility Principle).

- Second, considering how much code someone looking at `FeedImageStruct` needs to read to fully understand it, and to understand how it's used inside our system. It's not clear at first glance that some decoding strategy is needed to properly use this model, and that information actually lives in another component if we use `keyDecodingStrategy`.

If there's a simpler solution where we better define the responsibilities, we should aim for that. Considering that `FeedImageStruct`'s sole reason is to describe the API models, I personally would go with what's easier to understand when reading, and that's my previous suggestion.

---

But another possibility to keep the responsibility inside `FeedImageStruct` is to define CodingKeys, so if we have to change in the future there's a single point of change regarding decoding strategy:
```suggestion
		var imageId: UUID
		var imageUrl: URL
		var imageDesc: String?
		var imageLoc: String?

		enum CodingKeys: String, CodingKey {
			case imageId = "image_id"
			case imageUrl = "image_url"
			case imageDesc = "image_desc"
			case imageLoc = "image_loc"
		}
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +30,30 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: self.url) { [weak self] result in
+			guard let _ = self else { return }
```
thiagopenna, line 34: Instead of using a `guard let` and ignoring the attribution, it's clearer to check for `nil`:
```suggestion
			guard self != nil else { return }
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +30,30 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+			guard httpResponse.statusCode == 200 else {
+				return completion(.failure(Error.invalidData))
+			}
+
+			let decoder = JSONDecoder()
+			decoder.keyDecodingStrategy = .convertFromSnakeCase
+			guard let response = try? decoder.decode(FeedImageResponse.self, from: data) else {
+				return completion(.failure(Error.invalidData))
+			}
```
thiagopenna, line 48: Since both of these statements complete with the same error, consider joining them into a single guard statement:
```suggestion
			let decoder = JSONDecoder()
			decoder.keyDecodingStrategy = .convertFromSnakeCase
			guard httpResponse.statusCode == 200,
			      let response = try? decoder.decode(FeedImageResponse.self, from: data) else {
				return completion(.failure(Error.invalidData))
			}
```

---

If you decide against using `keyDecodingStrategy` (following from my previous comment suggesting `CodingKeys` instead), it's possible to inline the `JSONDecoder` initialization:
```suggestion
			guard httpResponse.statusCode == 200,
			      let response = try? JSONDecoder().decode(FeedImageResponse.self, from: data) else {
				return completion(.failure(Error.invalidData))
			}
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +30,30 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+			let feedImages = response.items.map {
+				FeedImage(id: $0.imageId,
+				          description: $0.imageDesc,
+				          location: $0.imageLoc,
+				          url: $0.imageUrl)
+			}
```
thiagopenna, line 55: Consider creating helpers to move `mapping` logic outside of the `load` function, to keep it simple and easier to read.

thiagopenna (commented):
Hi @rph8, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
rph8: OK, don't really agree that this improves readability of the code here, but fine.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -18,7 +36,31 @@ public final class RemoteFeedLoader: FeedLoader {
...
 	}
 
+	private func feedImages(for response: FeedImageResponse) -> [FeedImage] {
+		return response.items.map {
+			FeedImage(id: $0.imageId,
+			          description: $0.imageDesc,
+			          location: $0.imageLoc,
+			          url: $0.imageUrl)
+		}
+	}
```
thiagopenna, line 46: Following from my previous comment, please consider separating all `mapping` logic from loading, as shown in the program. One suggestion is the following function:
```swift
private static func map(_ data: Data, from httpResponse: HTTPURLResponse) -> FeedLoader.Result
```

That way, after checking for `connectivity` errors, we can keep our `load` function short and concise by calling:
```swift
	completion(RemoteFeedLoader.map(data, from: httpResponse))
```
And the `map` function can be responsible for returning `.failure(Error.invalidData)` or `.success(feedImages)`. Also, feel free to create a separate `Mapper` class to deal with all tasks related to mapping both the httpResponse and data, to reduce the number of responsibilities from `RemoteFeedLoader`.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -5,6 +5,24 @@
...
+
+		enum CodingKeys: String, CodingKey {
+			case imageId = "image_id"
+			case imageUrl = "image_url"
+			case imageDesc = "image_desc"
+			case imageLoc = "image_loc"
+		}
+	}
+
+	struct FeedImageResponse: Codable {
```
thiagopenna, line 22: `FeedImageResponse` is an implementation detail of `RemoteFeedLoader`, and is not supposed to be accessible from outside this class. Please mark it as `private`:

Also, `Codable` is a type alias for the `Encodable` and `Decodable` protocols. In this case, we only need to decode `RemoteFeedImage` from an external representation so `Codable` can be replaced with `Decodable`.
```suggestion
	private struct FeedImageResponse: Decodable {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -5,6 +5,24 @@
 import Foundation
 
 public final class RemoteFeedLoader: FeedLoader {
+	struct FeedImageStruct: Codable {
```
thiagopenna, line 8: Following from my previous comment, the same applies to `FeedImageStruct`. The way it's being used at this moment it should be marked as `private`, but if you decide to reference this type in an extension (from one of my previous suggestions), you might also consider `fileprivate`.
```suggestion
	private struct FeedImageStruct: Decodable {
```

thiagopenna (commented):
Hi @rph8, you almost passed the challenge!

Thank you 🙂

thiagopenna:
Hi @rph8. Do you need any help performing the changes requested in the review? Just let me know - We're here to help! 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +34,28 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+			guard httpResponse.statusCode == 200 else {
+				return completion(.failure(Error.invalidData))
+			}
+
+			completion(RemoteFeedLoader.map(data, from: httpResponse))
+		}
+	}
+
+	private static func map(_ data: Data, from httpResponse: HTTPURLResponse) -> FeedLoader.Result {
```
thiagopenna, line 52: This function does not currently use the `httpResponse` parameter, so it should be removed.
Another option would be moving the `statusCode` verification inside of `map`, and keep the `httpResponse` parameter.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +34,28 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		let decoder = JSONDecoder()
+		decoder.keyDecodingStrategy = .convertFromSnakeCase
```
thiagopenna, line 54: Please refer to my previous [comment](https://github.com/essentialdevelopercom/ios-lead-essentials-feed-api-challenge/pull/265#discussion_r692061051) suggesting using `CodingKeys`, and let me know if you would consider moving the key decoding strategy responsibility to `FeedImageStruct` based on those reasons.

thiagopenna (commented):
Hi @rph8, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
rph8: Good point, thanks for pointing that out.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
rph8: Moving the responsibility to FeedImageResponse.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +40,26 @@ public final class RemoteFeedLoader: FeedLoader {
...
+	}
+
+	private static func map(_ data: Data, from httpResponse: HTTPURLResponse) -> FeedLoader.Result {
+		guard httpResponse.statusCode == 200 else {
+			return .failure(Error.invalidData)
+		}
+
+		guard let response = FeedImageResponse.decode(from: data) else {
+			return .failure(Error.invalidData)
+		}
```
thiagopenna, line 61: Since both `guard` statements return with the same type of `failure`, consider joining them:
```suggestion
		guard httpResponse.statusCode == 200,
		      let response = FeedImageResponse.decode(from: data) else {
			return .failure(Error.invalidData)
		}
```

thiagopenna (commented):
Hi @rph8,

---

PR #266, opened 2021-08-05
Stanislav Li - Feed API Challenge

gitlees:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -4,9 +4,20 @@
 
 import Foundation
 
+public struct FeedImageListRemote: Decodable {
```
danillahtin, line 7: `FeedImageListRemote` is a private implementation detail and it should not be visible from outside of the module. Moreover there's no need to make it internal as this type is only accessed inside `RemoteFeedLoader.swift`. Thus it can be private
Please make `FeedImageListRemote` private

```suggestion
private struct FeedImageListRemote: Decodable {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -4,9 +4,20 @@
...
+	public let items: [FeedImageRemote]
```
danillahtin, line 8: Since `FeedImageListRemote` should be private, no need to make its property public, please remove `public` modifier
```suggestion
	let items: [FeedImageRemote]
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -4,9 +4,20 @@
...
+}
+
+public struct FeedImageRemote: Decodable {
```
danillahtin, line 11: `FeedImageRemote` is a private implementation detail and it should not be visible from outside of the module. Moreover there's no need to make it internal as this type is only accessed inside `RemoteFeedLoader.swift`. Thus it can be private
Please make `FeedImageRemote` private

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -4,9 +4,20 @@
...
+	public let imageId: UUID
+	public let imageDesc: String?
+	public let imageLoc: String?
+	public let imageUrl: URL
```
danillahtin, line 15: Since `FeedImageRemote ` should be private, no need to make its properties public, please remove `public` modifiers

```suggestion
	let imageId: UUID
	let imageDesc: String?
	let imageLoc: String?
	let imageUrl: URL
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -4,9 +4,20 @@
...
+}
+
 public final class RemoteFeedLoader: FeedLoader {
 	private let url: URL
-	private let client: HTTPClient
+	private var client: HTTPClient
```
danillahtin, line 20: This code should not be changed, please revert the change

```suggestion
	private let client: HTTPClient
```

danillahtin (commented):
Hi @gitlees, you almost passed the challenge!

Thank you 🙂

gitlees:
@danillahtin Made the required changes. Please review further

Thank you

danillahtin (commented):
Hi @gitlees,

Great job! ✅

PR #268, opened 2021-08-07
Arsalan Akhtar - Feed API Challenge.

salanswift:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,51 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				completion(FeedItemMapper.map(data, response: response))
+
+			case .failure:
+				completion(.failure(Error.connectivity))
+			}
+		}
+	}
+}
+
+internal final class FeedItemMapper {
```
thiagopenna, line 38: `FeedItemMapper` is an implementation detail of `RemoteFeedLoader`, and since it's in the same file in should be marked as `private`:
```suggestion
private final class FeedItemMapper {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,51 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		let image_desc: String?
+		let image_loc: String?
+		let image_url: URL
+
+		var item: FeedImage {
+			return FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
+		}
+	}
+
+	static var OK_200: Int { return 200 }
```
thiagopenna, line 60: `OK_200` is an implementation detail of `FeedItemMapper`, and should be marked as `private`:
```suggestion
	private static var OK_200: Int { return 200 }
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,51 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+	internal static func map(_ data: Data, response: HTTPURLResponse) -> RemoteFeedLoader.Result {
```
thiagopenna, line 62: The convention in this codebase is to omit the default access level keyword `internal`, please remove it:
```suggestion
	static func map(_ data: Data, response: HTTPURLResponse) -> RemoteFeedLoader.Result {
```

Tests/LoadFeedFromRemoteUseCaseTests.swift
```diff
@@ -24,108 +24,115 @@ class LoadFeedFromRemoteUseCaseTests: XCTestCase {
...
-//	func test_loadTwice_requestsDataFromURLTwice() {
-//		let url = URL(string: "https://a-given-url.com")!
-//		let (sut, client) = makeSUT(url: url)
-//
-//		sut.load { _ in }
-//		sut.load { _ in }
-//
-//		XCTAssertEqual(client.requestedURLs, [url, url])
-//	}
-//
```
thiagopenna, line 36: Please remove extraneous commented lines in this file since they make the `diff` harder to read.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,51 @@ public final class RemoteFeedLoader: FeedLoader {
...
+}
+
+internal final class FeedItemMapper {
+	private struct Root: Decodable {
+		let items: [Item]
+
+		var feed: [FeedImage] {
+			return items.map {
+				$0.item
+			}
```
thiagopenna, line 45: Suggestion: it's possible to use `map` with key paths to achieve the same result:
```suggestion
			return items.map(\.item)
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
thiagopenna: `FeedItemMapper` is defined as a class, so we can instantiate it as `FeedItemMapper()`. However, all its methods are static - it has no instance methods or properties.

As you'll see in the following lectures, you can make the initializer private in such cases since the `FeedItemMapper` is just a namespace for static methods, with no instance-specific behavior.

Both solutions will disallow instantiation at compile time so that `FeedItemMapper()` won't compile ✅

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,51 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			guard self != nil else { return }
+
+			switch result {
+			case let .success((data, response)):
+
+				completion(FeedItemMapper.map(data, response: response))
+
+			case .failure:
+				completion(.failure(Error.connectivity))
+			}
```
thiagopenna, line 33: Please remove unnecessary extra lines. Suggestion:
```suggestion
			guard self != nil else { return }

			switch result {
			case let .success((data, response)):
				completion(FeedItemMapper.map(data, response: response))
			case .failure:
				completion(.failure(Error.connectivity))
			}
```

thiagopenna (commented):
Hi @salanswift ,

Great job! ✅

PR #269, opened 2021-08-08
Mher Movsisyan - Feed API Challenge

Movsisyan:


FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,52 @@
...
+	struct Root: Decodable {
+		private let items: [Item]
+
+		var feedItems: [FeedImage] {
+			items.compactMap { item in
+				guard let uuid = UUID(uuidString: item.imageId),
+							let url = URL(string: item.imageURL) else { return nil }
+
+				return FeedImage(id: uuid, description: item.imageDesc, location: item.imageLoc, url: url)
+			}
```
thiagopenna, line 21: Following from my previous comments, by taking advantage of the Type System there will be no need to cast the values to `UUID` and `URL`: JSONDecoder will take care of that.

Also, by returning `nil` this portion of code is simply ignoring invalid items. But in the case where one item cannot be mapped the whole result should be discarded and `RemoteFeedLoader` should complete with `invalidData`.

Since we don't need to deal with `nil` values by using the correct types, there's no need for the `guard` statement nor for the `compactMap`, so this portion can be simpler:
```suggestion
			items.map { item in
				return FeedImage(id: item.image_id, description: item.image_desc, location: item.image_loc, url: item.image_url)
			}
```

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,52 @@
...
+//  FeedAPIChallenge
+//
+//  Created by Mher Movsisyan on 08.08.21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+struct FeedItemsMapper {
+	struct Root: Decodable {
```
thiagopenna, line 12: `Root` is an implementation detail of `FeedItemsMapper`, so it should be marked as `private`:
```suggestion
	private struct Root: Decodable {
```

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,52 @@
...
+			items.compactMap { item in
+				guard let uuid = UUID(uuidString: item.imageId),
+							let url = URL(string: item.imageURL) else { return nil }
+
+				return FeedImage(id: uuid, description: item.imageDesc, location: item.imageLoc, url: url)
+			}
+		}
+	}
+
+	struct Item: Decodable {
```
thiagopenna, line 25: `Item` is an implementation detail of `FeedItemsMapper`, so it should be marked as `private`:
```suggestion
	private struct Item: Decodable {
```

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,52 @@
...
+	static let OK_200 = 200
+
+	static func map(data: Data, urlResponse: HTTPURLResponse) -> FeedLoader.Result {
+		if urlResponse.statusCode != FeedItemsMapper.OK_200 {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
+
+		guard let root = try? JSONDecoder().decode(Root.self, from: data) else {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
```
thiagopenna, line 48: It's possible to combine both the `if` and `guard` statements, since both paths return the same `invalidData` error:
```suggestion
		guard urlResponse.statusCode == OK_200,
		      let root = try? JSONDecoder().decode(Root.self, from: data) else {
			return .failure(RemoteFeedLoader.Error.invalidData)
		}
```
Also, `FeedItemsMapper.OK_200` can be simplified to `OK_200`, since this portion of code is already in a static context.

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,52 @@
...
+
+		enum CodingKeys: String, CodingKey {
+			case imageId = "image_id"
+			case imageURL = "image_url"
+			case imageDesc = "image_desc"
+			case imageLoc = "image_loc"
+		}
+	}
+
+	static let OK_200 = 200
```
thiagopenna, line 39: `OK_200` is an implementation detail of `FeedItemsMapper`, so it should be marked as `private`:
```suggestion
	private static let OK_200 = 200
```

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,52 @@
...
+		let imageURL: String
+		var imageDesc: String?
+		var imageLoc: String?
+
+		enum CodingKeys: String, CodingKey {
+			case imageId = "image_id"
+			case imageURL = "image_url"
+			case imageDesc = "image_desc"
+			case imageLoc = "image_loc"
+		}
```
thiagopenna, line 36: As shown in the lectures, the API data model properties can be named exactly as the Feed API specs/contract without the CodingKeys. This way, we can model the precise backend Feed API domain. 

```suggestion
```

thiagopenna (commented):
Hi @Movsisyan, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
Movsisyan: Done

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
Movsisyan: Done

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,52 @@
...
+				guard let uuid = UUID(uuidString: item.imageId),
+							let url = URL(string: item.imageURL) else { return nil }
+
+				return FeedImage(id: uuid, description: item.imageDesc, location: item.imageLoc, url: url)
+			}
+		}
+	}
+
+	struct Item: Decodable {
+		let imageId: String
```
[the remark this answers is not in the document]
Movsisyan: Great suggestion

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,52 @@
...
+		let imageURL: String
```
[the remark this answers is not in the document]
Movsisyan: Done

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
Movsisyan: Done

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
Movsisyan: Done

thiagopenna (commented):
Hi @Movsisyan,

Great job! ✅

PR #270, opened 2021-08-10
Jiri Urbasek - Feed API Challenge

ibru:
This PR implements the functionality of loading feed images by `RemoteFeedLoader`.

### DIFF
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -19,6 +19,6 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url, completion: { _ in })
 	}
 }
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -19,6 +19,17 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		client.get(from: url, completion: { _ in completion(.failure(Error.connectivity)) })
+		client.get(from: url, completion: { result in
+			do {
+				let successResult = try result.get()
+
+				guard successResult.1.statusCode == 200 else {
+					completion(.failure(Error.invalidData))
+					return
+				}
+			} catch {
+				completion(.failure(Error.connectivity))
+			}
+		})
 	}
 }
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -28,18 +28,17 @@ public final class RemoteFeedLoader: FeedLoader {
 			}
 
 			do {
-				let successResult = try result.get()
+				let (data, urlResponse) = try result.get()
 
-				guard successResult.1.statusCode == 200 else {
-					completionResult = .failure(Error.invalidData)
-					return
-				}
+				guard urlResponse.statusCode == 200 else { throw Error.invalidData }
 
-				let response = try JSONDecoder().decode(FeedItemsResponse.self, from: successResult.0)
-				let items = response.items.map(FeedImage.init)
+				let feedResponse = try JSONDecoder().decode(FeedItemsResponse.self, from: data)
+				let items = feedResponse.items.map(FeedImage.init)
 
 				completionResult = .success(items)
 
+			} catch let error as Error {
+				completionResult = .failure(error)
 			} catch is DecodingError {
 				completionResult = .failure(Error.invalidData)
 			} catch {
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -19,6 +19,6 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		client.get(from: url, completion: { _ in })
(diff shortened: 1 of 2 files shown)


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,60 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+				completionResult = .success(items)
+
+			} catch let error as Error {
+				completionResult = .failure(error)
+			} catch is DecodingError {
+				completionResult = .failure(Error.invalidData)
+			} catch {
+				completionResult = .failure(Error.connectivity)
+			}
```
thiagopenna, line 46: Instead of having a single `do/catch` block that deals with all types of errors, it's better to keep the logic simple and easy to read. This portion of code it actually dealing with 3 error paths, from 2 different logics: `loading` and `mapping`.

Consider splitting part of this logic into a helper method to deal with `mapping` logic, so the `load` function gets easier to follow.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,60 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			defer {
+				completion(completionResult)
+			}
+
+			do {
+				let (data, urlResponse) = try result.get()
+
+				guard urlResponse.statusCode == 200 else { throw Error.invalidData }
+
+				let feedResponse = try JSONDecoder().decode(FeedItemsResponse.self, from: data)
```
thiagopenna, line 35: Since a `JSONDecoder` error would also lead to `Error.invalidData`, it's possible to combine these two statements into a single one, to simplify our logic:
```suggestion
				guard urlResponse.statusCode == 200,
					  let feedResponse = try? JSONDecoder().decode(FeedItemsResponse.self, from: data) else {
					throw Error.invalidData
				}
```
That way, we can remove the `catch is DecodingError` pattern from our `catches`.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,60 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		})
+	}
+}
+
+private struct FeedItemsResponse {
+	let items: [FeedImageDTO]
+}
+
+extension FeedItemsResponse: Codable {
+	struct FeedImageDTO: Codable {
```
thiagopenna, line 56: Same as my previous comment:
```suggestion
	struct FeedImageDTO: Decodable {
```

thiagopenna (commented):
Hi @ibru, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
ibru: Totally makes sense, changed it in 15c3409

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,60 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			}
+		})
+	}
+}
+
+private struct FeedItemsResponse {
+	let items: [FeedImageDTO]
+}
+
+extension FeedItemsResponse: Codable {
```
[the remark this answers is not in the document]
ibru: Totally makes sense, changed it in 15c3409

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,60 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		let imageDescription: String?
+		let imageLocation: String?
+		let imageURL: URL
+
+		enum CodingKeys: String, CodingKey {
+			case imageId = "image_id"
+			case imageDescription = "image_desc"
+			case imageLocation = "image_loc"
+			case imageURL = "image_url"
+		}
```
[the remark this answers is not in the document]
ibru: Thanks, I like the idea. I've changed it in 9fdbad3

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,60 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url, completion: { [weak self] result in
+			guard self != nil else { return }
+
+			var completionResult: FeedLoader.Result = .failure(Error.invalidData)
```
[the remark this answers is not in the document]
ibru: Completely right. I ended up removing `defer` block entirely in 858c07f

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,60 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			defer {
+				completion(completionResult)
+			}
```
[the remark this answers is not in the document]
ibru: Thanks for good suggestion.

I like to declare `let completionResult` in the beginning and make complier check for me that it gets assigned a value and that it get's assigned only once.
This substitues pretty much the thing I wanted to achieve with `defer` block and is more readable.

I like to keep `completionResult` instead of calling `completion(..)` directly because it would then be more vulnerable to forget to call it or call twice.

Changed in 858c07f

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
ibru: I've split part into private method in 1b9b955

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
ibru: Simplified and separated this logic into separate method in 1b9b955

PR #275, opened 2021-08-14
Cecil Gironde - Feed API Challenge

cegrnd:
tests are passing

### DIFF
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoaderMapper.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoaderMapper.swift
@@ -1,0 +1,40 @@
+
+import Foundation
+
+internal final class RemoteFeedImagesMapper {
+	private struct Root: Decodable {
+		let remoteFeedImages: [RemoteFeedImage]
+
+		enum CodingKeys: String, CodingKey {
+			case remoteFeedImages = "items"
+		}
+
+		var feedImages: [FeedImage] {
+			return remoteFeedImages.map { $0.remoteFeedImage }
+		}
+	}
+
+	private struct RemoteFeedImage: Decodable {
+		let image_id: UUID
+		let image_desc: String?
+		let image_loc: String?
+		let image_url: URL
+
+		var remoteFeedImage: FeedImage {
+			return FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
+		}
+	}
+
+	private static var OK_200: Int { return 200 }
+
+	internal static func map(_ data: Data, from response: HTTPURLResponse) -> RemoteFeedLoader.Result {
+		guard response.statusCode == OK_200 else {
+			// HTTP is a transport layer so any error indicates the payload could not be delivered to the application
+			return .failure(RemoteFeedLoader.Error.connectivity)
+		}
+		guard let root = try? JSONDecoder().decode(Root.self, from: data) else {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
+		return .success(root.feedImages)
+	}
+}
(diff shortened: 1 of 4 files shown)


FeedAPIChallenge/Feed API/RemoteFeedLoaderMapper.swift
```diff
@@ -0,0 +1,40 @@
...
+		var remoteFeedImage: FeedImage {
+			return FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
+		}
+	}
+
+	private static var OK_200: Int { return 200 }
+
+	internal static func map(_ data: Data, from response: HTTPURLResponse) -> RemoteFeedLoader.Result {
+		guard response.statusCode == OK_200 else {
+			// HTTP is a transport layer so any error indicates the payload could not be delivered to the application
```
caiozullo, line 32: Hi @cegrnd!

As explained in the program, connectivity errors in this use case are only for failure to connect between client and server.

In the non-200 case, the connection worked out fine, and we received a response. However, it was not the response we were expecting based on the API contract. 

So it's not a connectivity error. It's an invalid data response.

Tests/LoadFeedFromRemoteUseCaseTests.swift
```diff
@@ -24,108 +24,108 @@ class LoadFeedFromRemoteUseCaseTests: XCTestCase {
...
+		})
+	}
+
+	func test_load_deliversInvalidDataErrorOnNon200HTTPResponse() {
+		let (sut, client) = makeSUT()
+
+		let samples = [199, 201, 300, 400, 500]
+
+		samples.enumerated().forEach { index, code in
+			expect(sut, toCompleteWith: .failure(.connectivity), when: {
```
caiozullo, line 52: To pass the challenge, the tests should not be changed.

The tests are here to validate the implementation. If they're failing, the implementation is wrong.

FeedAPIChallenge/Feed API/RemoteFeedLoaderMapper.swift
```diff
@@ -0,0 +1,40 @@
+
+import Foundation
+
+internal final class RemoteFeedImagesMapper {
+	private struct Root: Decodable {
+		let remoteFeedImages: [RemoteFeedImage]
+
+		enum CodingKeys: String, CodingKey {
+			case remoteFeedImages = "items"
+		}
```
caiozullo, line 10: CodingKeys is not needed in this context if you name the property `items`.

This way, the code is shorter and more expressive:

```suggestion
		let items: [RemoteFeedImage]
```

FeedAPIChallenge/Feed API/RemoteFeedLoaderMapper.swift
```diff
@@ -0,0 +1,40 @@
+
+import Foundation
+
+internal final class RemoteFeedImagesMapper {
```
caiozullo, line 4: `internal` is the default access level so you can omit it if you want to (we type it explicitly in the lectures for teaching purposes, but you don't need to type it in your code).

You can remove the `internal` keyword from anywhere used.

```suggestion
final class RemoteFeedImagesMapper {
```

caiozullo (commented):
Hi @cegrnd, you almost passed the challenge!

Thank you 🙂

---

P.S. I noticed you implemented the whole challenge in a single commit. I recommend you break it down into smaller commits going forward, so you can work in small batches and have working checkpoints you can easily revert to if needed. You can learn more in the readme file.

FeedAPIChallenge/Feed API/RemoteFeedLoaderMapper.swift
```diff
@@ -0,0 +1,33 @@
...
```
caiozullo, line 4: `internal` is the default access level so you can omit it if you want to (we type it explicitly in the lectures for teaching purposes, but you don't need to type it in your code).

You can remove the `internal` keyword from anywhere used.

```suggestion
final class RemoteFeedImagesMapper {
```

caiozullo (commented):
Hi @cegrnd,

I left a final review comment. Feel free to push changes.

PR #278, opened 2021-08-15
Andrew Marmion - Feed API Challenge

andrewmarmion:


FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,54 @@
...
+			)
+		}
+
+		enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case image = "image_url"
+		}
+	}
```
mapostolakis, line 41: As shown in the lectures, the API data model properties can be named exactly as the Feed API specs/contract without the CodingKeys. This way, we can model the precise backend Feed API domain.

But that's just a suggestion. If you and your team prefer using `CodingKeys`, that's fine too!

mapostolakis (commented):
Hi @andrewmarmion,

I left some final review comments. Feel free to push changes.

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
andrewmarmion: That makes sense. I have updated the `FeedImagesMapper` so that its `Item` no longer uses CodingKeys and the variables match the backend's convention.

mapostolakis: Nice one!

PR #279, opened 2021-08-16
Taras Zinchenko - Feed API Challenge

Karamzda:


FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,42 @@
...
+//  FeedImagesMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Taras Zinchenko on 16.08.2021.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+class FeedImagesMapper {
```
mapostolakis, line 11: This class is not intended to be subclassed so it can be marked as `final`.

```suggestion
final class FeedImagesMapper {
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,42 @@
...
+		let location: String?
+		let url: URL
+
+		private enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
+	}
```
mapostolakis, line 24: As shown in the lectures, the API data model properties can be named exactly as the Feed API specs/contract without the CodingKeys. This way, we can model the precise backend Feed API domain.

But that's just a suggestion. If you and your team prefer using `CodingKeys`, that's fine too!

mapostolakis (commented):
Hi @Karamzda,

I left some final review comments. Feel free to push changes.

PR #283, opened 2021-08-20
Tea Jeremic - Feed API Challenge

TessJeremic:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,12 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] response in
+			guard self != nil else { return }
+			switch response {
+			case .success(let result): completion(FeedImageMapper.map(data: result.0, from: result.1))
```
danillahtin, line 25: Suggestion:
`result` is a tuple and thus it can be decomposed right in the case description. It will make the code a little bit more descriptive and easier to read. Otherwise one probably will go to `FeedLoader` and try to figure out what the `result`'s type is.

```suggestion
			case let .success((data, response)): completion(FeedImageMapper.map(data: data, from: response))
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,38 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Tea Jeremic on 19.8.21..
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedImageMapper {
```

```suggestion
final class FeedImageMapper {
```

Suggestion:

```suggestion
enum FeedImageMapper {
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,38 @@
...
+		let image_desc: String?
+		let image_loc: String?
+		let image_url: URL
+
+		var feedImage: FeedImage {
+			return FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
+		}
+	}
+
+	static private let ValidStatusCode = 200
```
danillahtin, line 30: Suggestion:
There're different practices how developers name constants. In swift it is more common to treat constants as properties and name it in camel case starting with a small letter. Some people prefer using capitalized snake case (THIS_IS_A_CONSTANT). 
At the same time naming a constant with camel case starting with a **capital** letter can confuse other developers as usually it is types that are name this way. To avoid this confusion it is better to use either **small** first letter or some other styling.

```suggestion
	static private let validStatusCode = 200
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,38 @@
...
+	internal static func map(data: Data, from response: HTTPURLResponse) -> FeedLoader.Result {
```
danillahtin, line 31: To follow up on my previous comment please remove `internal` modifier
```suggestion
	static func map(data: Data, from response: HTTPURLResponse) -> FeedLoader.Result {
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,38 @@
...
+//
+//  Created by Tea Jeremic on 19.8.21..
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedImageMapper {
+	private struct RemoteObject: Decodable {
+		var items: [RemoteFeedImage]
```
danillahtin, line 13: Suggestion:
This property is never mutate so it can be `let` instead of `var`
```suggestion
		let items: [RemoteFeedImage]
```

danillahtin (commented):
Hi @TessJeremic, you almost passed the challenge!

Thank you 🙂

TessJeremic:
Hi Danil, I've made suggested changes

danillahtin (commented):
Hi @TessJeremic,

PR #284, opened 2021-08-21
Challenge 1 Feed API

klaudiusivan:


FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,42 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Klaudius Ivan on 21/08/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedImageMapper {
```

```suggestion
final class FeedImageMapper {
```

Suggestion:

```suggestion
enum FeedImageMapper {
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	internal static func map(_ data: Data, _ response: HTTPURLResponse) throws -> [FeedImage] {
+		guard response.statusCode == 200 else {
+			throw RemoteFeedLoader.Error.invalidData
+		}
+
+		let root = try JSONDecoder().decode(Root.self, from: data)
+		return root.items.map({ $0.item })
+	}
```
danillahtin, line 41: Please remove explicit `internal` in line 34

Suggestion:
`FeedItemsMapper.map` and `RemoteFeedLoader.map` are parts of one mapping process. They together map data and response to `FeedLoader.Result`. They can be separated in a way that it is now but there's no need to do this at the moment since `FeedItemsMapper` is internal and not used from anywhere except for `RemoteFeedLoader`.
Thus this two functions can be combined into one.
It will remove duplication of `Error.invalidData` in both functions and make the code shorter.

```suggestion
	static func map(_ data: Data, _ response: HTTPURLResponse) throws -> FeedLoader.Result {
		guard response.statusCode == 200, let root = try? JSONDecoder().decode(Root.self, from: data) else {
			return .failure(RemoteFeedLoader.Error.invalidData)
		}

		return .success(root.items.map({ $0.item }))
	}
```

danillahtin (commented):
Hi @klaudiusivan, you almost passed the challenge!

Thank you 🙂

PR #285, opened 2021-08-22
Wayne Hsiao - Feed API Challenge

wayneshiau2002:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -4,6 +4,37 @@
 
 import Foundation
 
+final class FeedImageMapper {
```
danillahtin, line 7: `FeedImageMapper` is not used from outside of the file thus it can be private, please make it `private`
```suggestion
private final class FeedImageMapper {
```

Suggestion:

```suggestion
private enum FeedImageMapper {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +50,15 @@ public final class RemoteFeedLoader: FeedLoader {
...
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
+
+			switch result {
+			case let .success((data, response)):
+				completion(FeedImageMapper.map(data, from: response))
+			case .failure(_):
```
danillahtin, line 59: `(_)` is not needed, please remove it:
```suggestion
			case .failure:
```

danillahtin (commented):
Hi @wayneshiau2002, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
wayneshiau2002: @danillahtin Got it. Really appreciate your instruction on making `final class` to `private enum`. I didn't realize the difference before.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
wayneshiau2002: @danillahtin I've changed these two things base on your instructions. Thank you.

danillahtin (commented):
Hi @wayneshiau2002,

PR #287, opened 2021-08-23
Frank Carnevale - Feed API Challenge

fcarnevale:


FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,49 @@
...
+//  FeedItemsMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Frank Carnevale on 8/23/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedItemsMapper {
```
danillahtin, line 11: `internal` is the default access level, so you don't need to write it explicitly.

```suggestion
final class FeedItemsMapper {
```

danillahtin (commented):
Hi @fcarnevale, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
fcarnevale: Done. Thanks!

danillahtin:
Hi @fcarnevale, nice one!

PR #298, opened 2021-08-28
Luis Garcia - Feed API Challenge

iamluisg:
Implements load method in RemoteFeedLoader file and implements all test cases in LoadFeedFromRemoteUseCaseTest file.


FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,48 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Luis Garcia on 8/27/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedImageMapper {
```

```suggestion
final class FeedImageMapper {
```

Suggestion:

```suggestion
enum FeedImageMapper {
```

danillahtin (commented):
Hi @iamluisg, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/FeedImageMapper.swift
iamluisg: Hello @danillahtin, thank you for your suggestions and review, I have pushed a new commit with changes. Great tip about the private init for a class that only contains static methods! One question, I know that internal is the default, but isn't there value in being explicit in some cases, especially when building frameworks? Thanks!

danillahtin: Hi @iamluisg,
Probably there's some value in explicit internals. 
What we think here in essential developer is that it is very important to be consistent and disciplined. If the codebase doesn't have explicit internals and somebody wants to introduce one then they should either introduce it everywhere or leave everything as is. Consistent codebase is easier to read, understand and thus maintain.
I don't say that explicit internal is good or bad. It is a decision that every developer or team should make.

danillahtin (commented):
Hi @iamluisg,

PR #299, opened 2021-08-28
Gerson Janhuel - Feed API Challenge

gersonjanhuel:
Hi Caio & Mike this is my submisson for Feed API Challenge

### DIFF
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -19,6 +19,6 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { _ in }
 	}
 }
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -25,41 +25,41 @@ public final class RemoteFeedLoader: FeedLoader {
 				guard response.statusCode == 200 else {
 					return completion(.failure(Error.invalidData))
 				}
-				
-				// decode process 
+
+				// decode process
 				if let root = try? JSONDecoder().decode(Root.self, from: data) {
 					completion(.success(root.feedImages))
 				} else {
 					return completion(.failure(Error.invalidData))
 				}
-				
+
 			default:
 				completion(.failure(Error.connectivity))
 			}
 		}
 	}
-	
+
 	private struct Root: Decodable {
 		var items: [Image]
-		
+
 		var feedImages: [FeedImage] {
 			return items.map { $0.feedImage }
 		}
 	}
-	
+
 	private struct Image: Decodable {
 		let id: UUID
 		let imageDesc: String?
 		let imageLoc: String?
 		let imageURL: URL
-		
+
 		enum CodingKeys: String, CodingKey {
 			case id = "image_id"
 			case imageDesc = "image_desc"
 			case imageLoc = "image_loc"
 			case imageURL = "image_url"
 		}
-		
+
 		var feedImage: FeedImage {
 			return FeedImage(id: id, description: imageDesc, location: imageLoc, url: imageURL)
 		}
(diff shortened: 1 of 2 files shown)


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,59 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			// to prevent memory leak
+			guard self != nil else { return }
```
danillahtin, line 24: This comment can be confusing. Now is it as a comment for line 24 and I don't think that `guard self != nil else { return }` prevents memory leak.
In general there's no need to leave such comments as all the behaviour is documented either with properly written code or with properly written tests.
Please remove the comment
```suggestion
		client.get(from: url) { [weak self] result in
			guard self != nil else { return }
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,59 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+			switch result {
+			case let .success((data, response)):
+				completion(FeedImagesMapper.map(data,response))
```
danillahtin, line 28: Please add spacing after `,` to follow the project's code style convention
```suggestion
				completion(FeedImagesMapper.map(data, response))
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,59 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+			default:
+				completion(.failure(Error.connectivity))
```
danillahtin, line 31: It can be confusing to see that a default case completes with a connectivity error. Also the developer has to keep in mind that `default` actually means `failure`.
There's no extra value in having `default` here and it make the code harder to understand. 
```suggestion
			.failure:
				completion(.failure(Error.connectivity))
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,59 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			}
+		}
+	}
+	
+	private final class FeedImagesMapper {
```
danillahtin, line 36: Suggestion:
`FeedImagesMapper` is defined as a class, so we can instantiate it as `FeedImagesMapper()`. However, all its methods are static - it has no instance methods or properties.

```suggestion
	private enum FeedImagesMapper {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,59 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		private struct Root: Decodable {
+			var items: [Image]
```
danillahtin, line 38: `items` is never changed thus it should be `let`
```suggestion
			let items: [Image]
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,59 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		static func map(_ data: Data, _ response: HTTPURLResponse) -> FeedLoader.Result {
+			guard response.statusCode == 200 else {
+				return .failure(Error.invalidData)
+			}
+			
+			if let root = try? JSONDecoder().decode(Root.self, from: data) {
+				return .success(root.feedImages)
+			} else {
+				return .failure(Error.invalidData)
+			}
```
danillahtin, line 72: Both failure cases can be combined in one `guard ...` statement to remove duplication
```suggestion
			guard response.statusCode == 200, let root = try? JSONDecoder().decode(Root.self, from: data) else {
				return .failure(Error.invalidData)
			}
			
			return .success(root.feedImages)			
```

danillahtin (commented):
Hi @gersonjanhuel, you almost passed the challenge!

Thank you 🙂

gersonjanhuel:
> Hi @gersonjanhuel, you almost passed the challenge!
> 
> 
> 
> Thank you 🙂

Hi Danil, appreciate your feedbacks. Those are awesome!
I've done several changes, please review it again. 
Thaaank you! 😁

danillahtin (commented):
Hi @gersonjanhuel,

PR #300, opened 2021-08-28
Arpad Zalan - Feed API Challenge

azalan:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,15 @@ public final class RemoteFeedLoader: FeedLoader {
...
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
+
+			switch result {
+			case let .success((data, response)):
+				completion(FeedImagesMapper.map(data, from: response))
+			default:
```
danillahtin, line 28: It can be confusing to see that a default case completes with a connectivity error. Also the developer has to keep in mind that `default` actually means `failure`.
There's no extra value in having `default` here and it makes the code harder to understand. 
```suggestion
			.failure:
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,46 @@
...
+//  FeedImagesMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Arpad Zalan on 2021. 08. 28.
+//  Copyright (c) 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedImagesMapper {
```
danillahtin, line 11: `internal` is the default access level, so you don't need to write it explicitly.

```suggestion
final class FeedImagesMapper {
```

Suggestion:
`FeedImagesMapper` is defined as a class, so we can instantiate it as `FeedImagesMapper()`. However, all its methods are static - it has no instance methods or properties.

```suggestion
enum FeedImagesMapper {
```

danillahtin (commented):
Hi @azalan, you almost passed the challenge!

Please, check my review comment and let me know once you're done so I can review it again. 

Thank you 🙂

azalan:
Hi @danillahtin, thanks for the review. I've updated the code according to your suggestions. Please take a look.

danillahtin (commented):
Hi @azalan,

PR #301, opened 2021-08-29
Paras Gorasiya - Feed API Challenge

prgorasiya:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,19 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			guard self != nil else { return }
+			
+			switch result {
+			case let .success((data, response)):
+				guard response.statusCode == 200, let images = try? RemoteFeedImageMapper.feedImages(from: data) else {
+					completion(.failure(Error.invalidData))
+					return
+				}
+				completion(.success(images))
+			case .failure(_):
```
danillahtin, line 32: `(_)` is not needed, please remove it:
```suggestion
			case .failure:
```

FeedAPIChallenge/Feed API/RemoteFeedImageMapper.swift
```diff
@@ -0,0 +1,35 @@
...
+//  RemoteFeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by paras gorasiya on 29/08/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+struct RemoteFeedImageMapper {
```
danillahtin, line 11: `RemoteFeedImageMapper` is defined as a struct, so we can instantiate it as `RemoteFeedImageMapper()`. However, all its methods are static - it has no instance methods or properties.

As you'll see in the following lectures, you can make the initializer private in such cases since the `RemoteFeedImageMapper` is just a namespace for static methods, with no instance-specific behavior.

Both solutions will disallow instantiation at compile time so that `RemoteFeedImageMapper()` won't compile ✅
```suggestion
enum RemoteFeedImageMapper {
```

danillahtin (commented):
Hi @prgorasiya,

---

FeedAPIChallenge/Feed API/RemoteFeedImageMapper.swift
prgorasiya: Thanks @danillahtin for your comments, I have made changes as per your instructions. Thanks

danillahtin:
Hi @prgorasiya,

Nice one! 👍

PR #302, opened 2021-08-29
Nagaraju - Feed API Challenge

varthalan:
Kindly help to review my PR.


FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,38 @@
+//
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+final class FeedImageMapper {
```
danillahtin, line 7: `FeedImageMapper` is defined as a class, so we can instantiate it as `FeedImageMapper()`. However, all its methods are static - it has no instance methods or properties.

```suggestion
enum FeedImageMapper {
```

danillahtin (commented):
Hi @varthalan,

---

FeedAPIChallenge/Feed API/FeedImageMapper.swift
varthalan: Thanks a lot Danil for helping to review this PR. Made the suggested change in `FeedImageMapper` and pushed change. Please help to review it.

danillahtin:
Hi @varthalan,

Nice one! 👍

varthalan:
> Hi @varthalan,
> 
> Nice one! 👍

Thanks alot Danil for reviewing this PR. Got to know that we can have an enum without cases. Thanks for the tip 🙏

PR #303, opened 2021-08-29
Vitaly Shpinyov - Feed API Challenge

vitas168:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,16 @@ public final class RemoteFeedLoader: FeedLoader {
...
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+
+			guard self != nil else { return }
+
+			switch result {
+			case let .success((data, response)):
+				completion(RemoteFeedImageMapper.map(data, from: response))
+			case .failure(_):
```
danillahtin, line 29: `(_)` is not needed, please remove it
```suggestion
			case .failure:
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,16 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				completion(.failure(RemoteFeedLoader.Error.connectivity))
```
danillahtin, line 30: `RemoteFeedLoader.` is not needed since this line is inside the scope, please remove `RemoteFeedLoader.`

```suggestion
				completion(.failure(Error.connectivity))
```

FeedAPIChallenge/Feed API/RemoteFeedImageMapper.swift
```diff
@@ -0,0 +1,43 @@
...
+//  RemoteFeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Vitaly Shpinyov on 29.08.2021.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class RemoteFeedImageMapper {
```
danillahtin, line 11: `internal` is the default access level, so you don't need to write it explicitly.

```suggestion
final class RemoteFeedImageMapper {
```

`RemoteFeedImageMapper` is defined as a class, so we can instantiate it as `RemoteFeedImageMapper()`. However, all its methods are static - it has no instance methods or properties.

As you'll see in the following lectures, you can make the initializer private in such cases since the `RemoteFeedImageMapper` is just a namespace for static methods, with no instance-specific behavior.

Both solutions will disallow instantiation at compile time so that `RemoteFeedImageMapper()` won't compile ✅
```suggestion
enum RemoteFeedImageMapper {
```

FeedAPIChallenge/Feed API/RemoteFeedImageMapper.swift
```diff
@@ -0,0 +1,43 @@
...
+			                 location: image_loc,
+			                 url: image_url)
+		}
+	}
+
+	private static var OK_200: Int { return 200 }
+
+	internal static func map(_ data: Data, from response: HTTPURLResponse) -> FeedLoader.Result {
+		guard response.statusCode == OK_200,
+		      let json = try? JSONDecoder().decode(Root.self, from: data)
```
danillahtin, line 37: Suggestion:
It is better to call it `root` instead of `json` because it is not a json and the name can be confusing
```suggestion
		      let root = try? JSONDecoder().decode(Root.self, from: data)
```

danillahtin (commented):
Hi @vitas168, you almost passed the challenge!

Thank you 🙂

vitas168:
> Hi @vitas168, you almost passed the challenge!
> 
> 
> 
> Thank you 🙂

Hey Danil,

Thanks a lot for your review comments. They're much to the point!

Please review the amended PR.

Cheers,
Vitaly

danillahtin (commented):
Hi @vitas168,

PR #304, opened 2021-08-30
Augusto Monteiro - Feed API Challenge

augustomontc:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,14 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
+			switch result {
+			case let .success((data, response)):
+				completion(FeedImageMapper.map(data, from: response))
+			default:
```
danillahtin, line 27: It can be confusing to see that a default case completes with a connectivity error. Also the developer has to keep in mind that `default` actually means `failure`.
There's no extra value in having `default` here and it makes the code harder to understand. 
```suggestion
			.failure:
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,38 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Augusto Monteiro on 29/08/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedImageMapper {
```

```suggestion
final class FeedImageMapper {
```

```suggestion
enum FeedImageMapper {
```

danillahtin (commented):
Hi @augustomontc, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
augustomontc: Oops, that was totally my mistake! Just a lack of attention. 😄 
Thank you for the feedback!

FeedAPIChallenge/Feed API/FeedImageMapper.swift
augustomontc: Yeah, in fact I never use those internal keywords in my projects. Just put it to follow the lectures as it is.
That's great! Thank you for the feedback!

augustomontc: For the second topic, I prefer to make the initializer private, since making it an enum may be confusing in the future. Do you see any benefits making it an enum instead?

augustomontc:
Hi @danillahtin! Thank you for the feedbacks! 
I left a question on the comments, if you could clarify it to me it would be awesome!

The PR is ready to be reviewed again! See ya!

FeedAPIChallenge/Feed API/FeedImageMapper.swift
danillahtin: > Do you see any benefits making it an enum instead?
Yes, it is less code. One can always forget to put `final` before `class` and to makethe initializer private. It looks like a boiler plate that can be solved with 4 letters - `enum` 😁
But everybody decides on their own. I agree that it can be confusing for the person who is not familiar with this trick.

danillahtin (commented):
Hi @augustomontc,

PR #305, opened 2021-08-30
Joshua Bryson - Feed API Challenge.

DrCarleeknikov:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,50 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
+			switch result {
+			case .success((let data, let response)):
+				completion(FeedItemsMapper.map(data, from: response))
+			case .failure(_):
```
thiagopenna, line 27: Since the error attribution is ignored, we can omit it:
```suggestion
			case .failure:
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,50 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+		private enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
+	}
+
+	internal static func map(_ data: Data, from response: HTTPURLResponse) -> RemoteFeedLoader.Result {
```
thiagopenna, line 61: The convention in this codebase is to omit the default access level keyword of `internal`, please remove it.
```suggestion
	static func map(_ data: Data, from response: HTTPURLResponse) -> RemoteFeedLoader.Result {
```

PR #306, opened 2021-08-31
Sergio Khalil Bello García - Feed API Challenge.

i-khalil-s:
Feed API Challenge complete!

### DIFF
--- a/FeedAPIChallenge/Feed API/FeedImageMapper.swift
+++ b/FeedAPIChallenge/Feed API/FeedImageMapper.swift
@@ -1,0 +1,29 @@
+//
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Sergio Khalil Bello Garcia on 30/08/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedImageMapper {
+	internal struct Root: Decodable {
+		let images: [Image]
+	}
+
+	internal struct Image: Decodable {
+		let image_id: UUID
+		let image_desc, image_loc: String?
+		let image_url: URL
+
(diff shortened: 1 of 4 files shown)


thiagopenna (commented):
Hi @i-khalil-s, you almost passed the challenge!

Thank you 🙂

i-khalil-s:
@thiagopenna Thanks for the feedback!

I've made the necessary corrections you suggested to keep the code in a good fit!

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,39 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Sergio Khalil Bello Garcia on 30/08/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+final class FeedImageMapper {
```
thiagopenna, line 11: `FeedImageMapper` is defined as a class, so we can instantiate it as `FeedImageMapper()`. However, all its methods are static - it has no instance methods or properties.

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,39 @@
...
+	private struct Root: Decodable {
+		let items: [Image]
+
+		var feed: [FeedImage] {
+			return items.map { $0.feedImage }
```
thiagopenna, line 16: Suggestion: another way of using the `map` function is with key paths:
```suggestion
			return items.map(\.feedImage)
```

thiagopenna (commented):
Hi @i-khalil-s,

---

thiagopenna:
Well done!

PR #309, opened 2021-09-01
Farzad Zamani - Feed API Challenge.

farzadzamani:


FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,51 @@
...
+			return FeedImage(id: imageId, description: imageDescription, location: imageLocation, url: imageURL)
+		}
+	}
+
+	static func map(_ data: Data, _ response: HTTPURLResponse) throws -> [FeedImage] {
+		guard response.OK_200 else {
+			throw RemoteFeedLoader.Error.invalidData
+		}
+
+		let root = try JSONDecoder().decode(Root.self, from: data)
```
thiagopenna, line 41: Instead of `throwing` `JSONDecoder` errors and have clients of this function map them to meaningful errors, we can perform that here, having a single path that `throws` `.invalidData`.
```suggestion
		guard response.OK_200,
		      let root = try? JSONDecoder().decode(Root.self, from: data) else {
			throw RemoteFeedLoader.Error.invalidData
		}
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,20 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+			guard self != nil else { return }
+
+			switch clientResult {
+			case .success((let data, let response)):
+				if let feedImages = try? FeedImageMapper.map(data, response) {
+					completion(.success(feedImages))
+				} else {
+					completion(.failure(Error.invalidData))
+				}
```
thiagopenna, line 32: Following from my previous comment, by having the `map` function `throw` the correct error we don't have to overwrite it here. That way we could use a `do/catch` statement and complete with the `error` caught.

```suggestion
				completion(Result { try FeedImageMapper.map(data, response) })
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,20 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		self.client.get(from: url) { [weak self] clientResult in
+
+			guard self != nil else { return }
+
+			switch clientResult {
+			case .success((let data, let response)):
```
thiagopenna, line 27: This success `case` can be simplified with a single let.
```suggestion
			case let .success((data, response)):
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,51 @@
...
+import Foundation
+
+// MARK: - FeedImageMapper
+final class FeedImageMapper {
+	private struct Root: Decodable {
+		let items: [JsonFeedImageModel]
+	}
+
+	// MARK: - JsonFeedImageModel
+	struct JsonFeedImageModel: Decodable {
```
thiagopenna, line 18: `JsonFeedImageModel` is an implementation detail of `FeedImageMapper`, so it should be marked as `private`:
```suggestion
	private struct JsonFeedImageModel: Decodable {
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,51 @@
...
+		}
+
+		let root = try JSONDecoder().decode(Root.self, from: data)
+		return root.items.map { $0.feedImage }
+	}
+}
+
+// MARK: - HTTPURLResponse
+extension HTTPURLResponse {
+	var OK_200: Bool {
```
thiagopenna, line 48: Consider naming this property to make it clear that it's a boolean. Suggestion:
```suggestion
	var isOK: Bool {
```
That way we would use it as `guard response.isOK` instead of `guard response.OK_200`, improving readability and avoiding leaking implementation details (like the specific `statusCode`, that might change in the future)

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
farzadzamani: @thiagopenna The Result init point is very useful, Thanks

farzadzamani:
> Hi @farzadzamani, you almost passed the challenge!
> 
> This is Penna, I am an instructor at Essential Developer.
> 
> 
> Thank you 🙂

Hi @thiagopenna ,
All changes is done, Please review again,

Thanks.

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,51 @@
...
+//  FeedAPIChallenge
+//
+//  Created by Farzad Zamani on 9/1/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+// MARK: - FeedImageMapper
+final class FeedImageMapper {
```
thiagopenna, line 12: `FeedImageMapper` is defined as a class, so we can instantiate it as `FeedImageMapper()`. However, all its methods are static - it has no instance methods or properties.

thiagopenna (commented):
Hi @farzadzamani,

---

PR #311, opened 2021-09-03
Barry Chen - Feed API Challenge

Barry0327:


FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,48 @@
...
+//  FeedImagesMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Chen Yi-Wei on 2021/9/3.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal struct FeedImagesMapper {
```
thiagopenna, line 11: The convention in this codebase is to omit the default access level keyword of `internal`, so it should be removed.
```suggestion
struct FeedImagesMapper {
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,48 @@
...
+		}
+
+		var feedImage: FeedImage {
+			FeedImage(id: id, description: description, location: location, url: url)
+		}
+	}
+
+	private static var validCode: Int { 200 }
+
+	internal static func map(_ data: Data, from response: HTTPURLResponse) -> FeedLoader.Result {
```
thiagopenna, line 40: The convention in this codebase is to omit the default access level keyword of `internal`, so it should be removed.
```suggestion
	static func map(_ data: Data, from response: HTTPURLResponse) -> FeedLoader.Result {
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,48 @@
...
+//
+
+import Foundation
+
+internal struct FeedImagesMapper {
+	private struct Root: Decodable {
+		let items: [Item]
+
+		var feedImages: [FeedImage] {
+			items.map { $0.feedImage }
```
thiagopenna, line 16: Suggestion: another way of using the `map` function is to use key paths:
```suggestion
			items.map(\.feedImage)
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
thiagopenna: `FeedImageMapper` is defined as a struct, so we can instantiate it as `FeedImageMapper()`. However, all its methods are static - it has no instance methods or properties.

Barry0327: Hi Penna,
thanks for your review,
it's really very helpful,
I was wondering why not use struct in this case in the lecture,
your suggestion answered my question,
thanks a lot!

thiagopenna (commented):
Well done!

PR #313, opened 2021-09-04
Sushant kumat - Feed API Challenge #313

sush4:


mapostolakis:
Hi @sush4, if your PR is ready for review please post a comment on the challenge page in the academy with the link to your PR as described in the README file. Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -7,6 +7,7 @@ import Foundation
 public final class RemoteFeedLoader: FeedLoader {
 	private let url: URL
 	private let client: HTTPClient
+	private var ok_200: Int { return 200 }
```
thiagopenna, line 10: `ok_200` could be a `static` variable so we don't need to reference `self` when using this value.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +20,40 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard let this = self else { return }
```
thiagopenna, line 24: Instead of using `this` it would be clearer to use `self` for this attribution.
```swift
	guard let self = self else { return }
```
But following from my previous comment there's no reason to keep a reference to `self`, so we can just check for `nil`:
```suggestion
			guard self != nil else { return }
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +20,40 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			switch result {
+			case .success((let data, let response)):
+				if response.statusCode == this.ok_200, let feedImages = try? JSONDecoder().decode(FeedImageParser.self, from: data).feedImages {
```
thiagopenna, line 27: Following from my previous comments, by making `ok_200` a static property we could access it without referencing `self`:
```suggestion
				if response.statusCode == RemoteFeedLoader.ok_200, let feedImages = try? JSONDecoder().decode(FeedImageParser.self, from: data).feedImages {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +20,40 @@ public final class RemoteFeedLoader: FeedLoader {
...
+					completion(.success(feedImages))
+				} else {
+					completion(.failure(RemoteFeedLoader.Error.invalidData))
+				}
+			case .failure:
+				completion(.failure(RemoteFeedLoader.Error.connectivity))
```
thiagopenna, line 33: Since this portion of code is already inside the `RemoteFeedLoader` scope, we can make the error shorter:
```suggestion
				completion(.failure(Error.connectivity))
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +20,40 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			}
+		}
+	}
+}
+
+private struct FeedImageParser: Decodable {
```
thiagopenna, line 39: `FeedImageParser` is misleading, since the main responsibility for this `struct` is to represent the root level of the `JSON` object returned by the API. By naming it `parser`, another developer might expect it to be responsible for mapping the `data` directly.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +20,40 @@ public final class RemoteFeedLoader: FeedLoader {
...
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard let this = self else { return }
+			switch result {
+			case .success((let data, let response)):
+				if response.statusCode == this.ok_200, let feedImages = try? JSONDecoder().decode(FeedImageParser.self, from: data).feedImages {
+					completion(.success(feedImages))
+				} else {
+					completion(.failure(RemoteFeedLoader.Error.invalidData))
+				}
```
thiagopenna, line 31: Consider moving this logic to a helper function, to keep the `load` function shorter and easier to read.

thiagopenna (commented):
Hi @sush4, you almost passed the challenge!

Thank you 🙂

sush4:
> Hi @sush4, you almost passed the challenge!
> 
> This is Penna, I am an instructor at Essential Developer.
> 
> 
> Thank you 🙂

Thanks for the pull request review. Updated the pull request as you commented.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +20,44 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			switch result {
+			case .success((let data, let response)):
+				self.mapSuccessResponse(data: data, response: response, completion: completion)
+			case .failure:
+				completion(.failure(Error.connectivity))
+			}
+		}
+	}
+
+	private func mapSuccessResponse(data: Data, response: HTTPURLResponse, completion: @escaping (FeedLoader.Result) -> Void) {
```
thiagopenna, line 34: This function can be `static` since there's no need to have an instance of `RemoteFeedLoader` to perform this logic. That way, we avoid needing a strong reference to `self` inside the `load` function:
```suggestion
	private static func mapSuccessResponse(data: Data, response: HTTPURLResponse, completion: @escaping (FeedLoader.Result) -> Void) {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +20,44 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard let `self` = self else { return }
```
thiagopenna, line 24: Since [Swift 4.2](https://github.com/apple/swift/pull/15306) there's no need to use backticks to upgrade `self` from a weak reference.

Also following from my previous comment, we don't need a reference to `self` if we make the `mapSuccessResponse` function `static`.

thiagopenna (commented):
Hi @sush4, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
sush4: Thanks, Updated!

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
sush4: Updated, One doubt, should we prefer `static` func over instance func for this kinda cases?

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +20,44 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
+			switch result {
+			case .success((let data, let response)):
```
thiagopenna, line 26: This success `case` can be simplified with a single let.
```suggestion
			case let .success((data, response)):
```

thiagopenna (commented):
Hi @sush4,

---

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
sush4: Updated! Thanks!

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
thiagopenna: It would depend on what's the logic being performed, and if we have any reasons to have a reference to an instance. 

If we are using instance-specific values, it might make sense to have an `instance` function, or we might prefer to pass those values as arguments in a `static` function, it would depend on the operation. 

If we are **altering** instance-specific properties, or have any type of side-effect in the instance, then I would recommend `instance` functions (instead of `inout` parameters on a function).

thiagopenna:
Well done!

PR #314, opened 2021-09-05
Juan Dorado Torres - Feed API Challenge

fdorado985:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,19 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		client.get(from: url) { [weak self] result in
+			if self == nil { return }
+			switch result {
+			case let .success((data, response)):
+				do {
+					let items = try FeedImagesMapper.map(data, response)
+					completion(.success(items))
+				} catch {
+					completion(.failure(Error.invalidData))
+				}
```
thiagopenna, line 31: Following from my previous comment, by having the `map` function `throw` the correct error we don't have to overwrite it here. That way we could use a `do/catch` statement and complete with the `error` caught.

```suggestion
				completion(Result { try FeedImageMapper.map(data, response) })
```

thiagopenna (commented):
Hi @ fdorado985, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
fdorado985: Thanks for the review, I have an small doubt about how is this `Result` type working, is it using the following initializer
`public init(catching body: () throws -> Success)`?

I decided to make another change, instead of `throw`, I send the expected `FeedLoader.Result` I didn't try it at the first time because autocompletion was not working with that, should be that a bug?, I write it anyway and it worked so, I used that approach now.

fdorado985: Ohhh gotcha! I was trying to make it work, but I missed the try keyword 👍

thiagopenna: Exactly @ fdorado985, the initializer takes a throwing closure as a parameter, and since it can `throw`, we need the `try` keyword to invoke it. And there's no problem returning a `Result`, that's a valid solution too. 

My comment was just to point in the direction that since Swift has ways of converting between `Result` and `throwing closures` (in the form of [`init(catching:)`](https://developer.apple.com/documentation/swift/result/3139399-init) and [`get()`](https://developer.apple.com/documentation/swift/result/3139397-get)), we should aim to use what's given by the language when possible, instead of rewriting the same logic.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,14 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			if self == nil { return }
+			switch result {
+			case let .success((data, response)):
+				completion(FeedImagesMapper.map(data, response))
+			case .failure(_):
```
thiagopenna, line 27: Since this portion of code ignores the error from the `.failure` case, we can ignore it completely:
```suggestion
			case .failure:
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,41 @@
...
+			case url = "image_url"
+		}
+	}
+
+	static func map(_ data: Data, _ response: HTTPURLResponse) -> FeedLoader.Result {
+		guard response.statusCode == 200, let root = try? JSONDecoder().decode(Root.self, from: data) else {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
+
+		return .success(root.items.map { $0.item })
```
thiagopenna, line 39: Suggestion: another way of using the `map` function is to use key paths:
```suggestion
		return .success(root.items.map(\.item))
```

thiagopenna (commented):
Hi @fdorado985,

---

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
fdorado985: This is really cool, I'll update to use `Result.init(catching:)`, to keep this other way #TIL, thanks!

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
fdorado985: 😮 This is something completely new, I'm concerned, how do we know we are able to use keypaths for a function? or is it just for these higher functions?

thiagopenna: There are many ways of using key-path expressions, especially considering that the `subscript(keyPath:)` is available on all types. From the [documentation](https://docs.swift.org/swift-book/ReferenceManual/Expressions.html#ID563), in this context I'm mainly referring to:

> _"You can use a key path expression in contexts where you would normally provide a function or closure. Specifically, you can use a key path expression whose root type is `SomeType` and whose path produces a value of type Value, instead of a function or closure of type `(SomeType) -> Value`."_
```swift
struct Task {
    var description: String
    var completed: Bool
}
var toDoList = [
    Task(description: "Practice ping-pong.", completed: false),
    Task(description: "Buy a pirate costume.", completed: true),
    Task(description: "Visit Boston in the Fall.", completed: false),
]

// Both approaches below are equivalent.
let descriptions = toDoList.filter(\.completed).map(\.description)
let descriptions2 = toDoList.filter { $0.completed }.map { $0.description }
```

thiagopenna:
Well done!

PR #315, opened 2021-09-05
Mauro Worobiej - Feed API Challenge

mauroworobiej:


FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,49 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Mauro Worobiej on 05/09/2021.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+class FeedImageMapper {
```
nodediggity, line 11: This class is not intended to be inherited thus it should be `final`.
```suggestion
final class FeedImageMapper {
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,49 @@
...
+			case location = "image_loc"
+			case url = "image_url"
+		}
+
+		var feedImage: FeedImage {
+			FeedImage(id: id, description: description, location: location, url: url)
+		}
+	}
+
+	static var OK_200: Int { 200 }
```
nodediggity, line 38: `OK_200` is an implementation detail of the `FeedImageMapper` so this property should be marked as `private`.
```suggestion
private static var OK_200: Int { 200 }
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,49 @@
...
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+class FeedImageMapper {
+	private struct Root: Decodable {
+		let items: [ItemModel]
+
+		var feedImage: [FeedImage] {
```
nodediggity, line 15: Consider naming this as a plural to indicate it is a collection as opposed to a single value.
```suggestion
		var feedImages: [FeedImage] {
```

nodediggity (approved):
Hi @mauroworobiej,

---

FeedAPIChallenge/Feed API/FeedImageMapper.swift
mauroworobiej: Nice catch!

FeedAPIChallenge/Feed API/FeedImageMapper.swift
mauroworobiej: got it, thanks!

FeedAPIChallenge/Feed API/FeedImageMapper.swift
mauroworobiej: Make sense, thanks!

nodediggity (approved):

PR #316, opened 2021-09-06
Hugo Silva Cantanhede - Feed API Challenge

hugosilvac:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,15 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
+
+			switch result {
+			case .success((let data, let response)):
```
danillahtin, line 26: The `let` can be defined outside the tuple so the expression can be shorter:
```suggestion
			case let .success((data, response)):
```

FeedAPIChallenge/Feed Feature/FeedImageMapper.swift
```diff
@@ -0,0 +1,47 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Hugo Cantanhede on 03/09/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedImageMapper {
```
danillahtin, line 11: `internal` is the default access level, so you don't need to write it explicitly.

```suggestion
final class FeedImageMapper {
```

Suggestion:

```suggestion
enum FeedImageMapper {
```

danillahtin (commented):
Hi @hugosilvac, you almost passed the challenge!

Thank you 🙂

PR #317, opened 2021-09-06
Angela Lin - Feed API Challenge

tigclaw:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,14 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
+			switch result {
+			case .success((let data, let response)):
```
danillahtin, line 25: The `let` can be defined outside the tuple so the expression can be shorter:
```suggestion
			case let .success((data, response)):
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,47 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Angela Lin on 9/6/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedImageMapper {
```

```suggestion
final class FeedImageMapper {
```
Suggestion:

```suggestion
enum FeedImageMapper {
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,47 @@
...
+	internal static func map(_ data: Data, _ response: HTTPURLResponse) -> FeedLoader.Result {
+		guard response.statusCode == OK_200 else {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
+		do {
+			let root = try JSONDecoder().decode(NetworkFeedImageContainer.self, from: data)
+			return .success(root.feedImages)
+		} catch {
+			return .failure(RemoteFeedLoader.Error.invalidData)
+		}
```
danillahtin, line 45: Since the error inside `catch` block is ignored optional `try?` can be used.
```swift
guard response.statusCode == OK_200 else {
}

guard let root = try? JSONDecoder().decode(NetworkFeedImageContainer.self, from: data) else {
}

```

And now it becomes clear that two guard blocks can be combined into one to remove duplication of `return .failure(RemoteFeedLoader.Error.invalidData)`:
```suggestion
		guard response.statusCode == OK_200, let root = try? JSONDecoder().decode(NetworkFeedImageContainer.self, from: data) else {
			return .failure(RemoteFeedLoader.Error.invalidData)
		}

		return .success(root.feedImages)
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,47 @@
...
+
+		private enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
+	}
+
+	private struct NetworkFeedImageContainer: Decodable {
```
danillahtin, line 26: Suggestion:
Since this struct is private, there's no need to give extra context in the name of the type. It can be simplified to just `Container`
```suggestion
	private struct Container: Decodable {
```

danillahtin (commented):
Hi @tigclaw, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
tigclaw: Cool, I'll have to remember this in the future.

tigclaw:

Hi Danil @danillahtin  - thanks for reviewing! I've addressed your comments in latest commit.

FeedAPIChallenge/Feed API/FeedImageMapper.swift
danillahtin: There's no standard way here. Using an enum is my own preference as it is very short and doesn't require explicitly making the initializer private. If you feel like using structs or classes with private initializers it is absolutely ok.

danillahtin (commented):
Hi @tigclaw,

PR #318, opened 2021-09-06
Jovanny Espinal - Feed API Challenge

JovannyEspinal:


FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
+
+internal struct FeedImageMapper {
```
danillahtin, line 3: `internal` is the default access level, so you don't need to write it explicitly.

```suggestion
struct FeedImageMapper {
```

Suggestion:
`FeedImageMapper` is defined as a struct, so we can instantiate it as `FeedImageMapper()`. However, all its methods are static - it has no instance methods or properties.

It may be confusing for other developers to be able to instantiate a struct but then not be able to do anything with it.

```suggestion
enum FeedImageMapper {
```

danillahtin (commented):
Hi @JovannyEspinal, you almost passed the challenge!

Thank you 🙂

danillahtin:
Hi @JovannyEspinal. Do you need any help performing the changes requested in the review? Just let me know - We're here to help! 🙂

JovannyEspinal:
Nope! Life got in the way. Will be back to it soon!

PR #319, opened 2021-09-07
Ihwan Dede - Feed API Challenge

IhwanID:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,15 @@ public final class RemoteFeedLoader: FeedLoader {
...
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
+
+			switch result {
+			case let .success((data, response)):
+				completion(FeedImageMapper.map(data: data, response: response))
+			case .failure(_):
```
danillahtin, line 28: Suggestion:
`(_)` is not needed and can be removed
```suggestion
			case .failure:
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,46 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Ihwan on 07/09/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+final class FeedImageMapper {
```
danillahtin, line 11: Suggestion:

```suggestion
enum FeedImageMapper {
```

danillahtin (commented):
Hi @IhwanID,

---

**P.S. I left a couple of final review comments. Feel free to push changes.**

danillahtin:
Hi @IhwanID, nice one! 👍

PR #321, opened 2021-09-09
Andre Bongartz - Feed API Challenge

bitsmakerde:


FeedAPIChallenge/Feed API/FeedImageItemMapper.swift
```diff
@@ -0,0 +1,39 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by André Bongartz on 09.09.21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal enum FeedImageItemMapper {
```
nodediggity, line 11: `internal` is the default access level, so you don't need to write it explicitly.

```suggestion
 enum FeedImageItemMapper {
```

FeedAPIChallenge/Feed API/FeedImageItemMapper.swift
```diff
@@ -0,0 +1,39 @@
...
+		var image_desc: String?
+		var image_loc: String?
+		var image_url: URL
+
+		var item: FeedImage {
+			return FeedImage(id: image_id, description: image_desc, location: image_loc, url: image_url)
+		}
+	}
+
+	internal static func map(_ data: Data, response: HTTPURLResponse) -> RemoteFeedLoader.Result {
```
nodediggity, line 32: `internal` can also be removed here as this is the default access level.

```suggestion
static func map(_ data: Data, response: HTTPURLResponse) -> RemoteFeedLoader.Result {
```

FeedAPIChallenge/Feed API/FeedImageItemMapper.swift
```diff
@@ -0,0 +1,39 @@
...
+		guard response.statusCode == 200, let dataAsJson = try? JSONDecoder().decode(Root.self, from: data) else {
```
nodediggity, line 33: It is good practice to avoid using numbers directly in source code without explanation - by defining a named constant in place of numeric literals we can provide more context on what is represented for the next developer.

```suggestion
		guard response.statusCode == OK_200, let dataAsJson = try? JSONDecoder().decode(Root.self, from: data) else {
```

The `OK_200` property can be a static property as there's no need for the `RemoteFeedLoader` to retrieve a constant value. 

```swift
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -13,12 +13,23 @@ public final class RemoteFeedLoader: FeedLoader {
 		case invalidData
 	}
 
+	public typealias Result = FeedLoader.Result
```
nodediggity, line 16: This alias is unused, you can remove it.

```suggestion
```

nodediggity (approved):
Hi @bitsmakerde,

---

PR #323, opened 2021-09-10
Fernando De La Rosa - Feed API Challenge

fer96:


FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,50 @@
...
+		let failureResult = FeedLoader.Result.failure(RemoteFeedLoader.Error.invalidData)
+		guard httpResponse.statusCode == 200,
+		      let feedImages = map(data) else { return failureResult }
+
+		return .success(feedImages)
+	}
+
+	private static func map(_ data: Data) -> [FeedImage]? {
+		guard let root = try? JSONDecoder().decode(Root.self, from: data) else { return nil }
+		return root.feedImages
```
thiagopenna, line 48: If this function returns an optional `[FeedImage]?`, there's no need for a guard statement: we can return the result of `try?`
```suggestion
		try? JSONDecoder().decode(Root.self, from: data).feedImages
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,50 @@
...
+
+		internal var item: FeedImage {
+			return FeedImage(id: id, description: description, location: location, url: url)
+		}
+	}
+
+	internal static func resolves(_ data: Data, with httpResponse: HTTPURLResponse) -> FeedLoader.Result {
+		let failureResult = FeedLoader.Result.failure(RemoteFeedLoader.Error.invalidData)
+		guard httpResponse.statusCode == 200,
+		      let feedImages = map(data) else { return failureResult }
```
thiagopenna, line 41: `failureResult` does not add extra meaning when compared to its type (`Result.failure`), and this variable might not be needed in the success path, so I recommend simply returning the failure inside the `else` clause instead of preemptively assigning it to a variable:
```suggestion
		guard httpResponse.statusCode == 200, let feedImages = map(data) else {
			return .failure(RemoteFeedLoader.Error.invalidData)
		}
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,14 @@ public final class RemoteFeedLoader: FeedLoader {
...
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] response in
+			guard self != nil else { return }
+			switch response {
+			case let .success((data, httpResponse)):
+				completion(FeedImageMapper.resolves(data, with: httpResponse))
+			case .failure:
+				completion(.failure(RemoteFeedLoader.Error.connectivity))
```
thiagopenna, line 28: We can shorten the error declaration since this portion of code is in the same scope as `RemoteFeedLoader`:
```suggestion
				completion(.failure(Error.connectivity))
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,50 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Fernando De La Rosa on 08/09/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedImageMapper {
```
thiagopenna, line 11: The convention in this codebase is to omit the default access level keyword of `internal`, please remove all instances of `internal` from this file.

thiagopenna (commented):
Hi @fer96, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/FeedImageMapper.swift
fer96: Done

FeedAPIChallenge/Feed API/FeedImageMapper.swift
fer96: Done

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
fer96: Done

FeedAPIChallenge/Feed API/FeedImageMapper.swift
fer96: I agree, I just used it to be more explicit, but yeah, that doesn't make sense.

fer96:
Hi @thiagopenna I resolved the comments, hope you can review it them :)

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,49 @@
...
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+final class FeedImageMapper {
+	private struct Root: Decodable {
+		private let items: [Image]
+
+		internal var feedImages: [FeedImage] {
```
thiagopenna, line 15: The convention in this codebase is to omit the default access level keyword `internal`, please remove it.
```suggestion
		var feedImages: [FeedImage] {
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,49 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Fernando De La Rosa on 08/09/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+final class FeedImageMapper {
```
thiagopenna, line 11: `FeedImageMapper` is defined as a class, so we can instantiate it as `FeedImageMapper()`. However, all its methods are static - it has no instance methods or properties.

thiagopenna (commented):
Hi @fer96,

---

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,49 @@
...
+	private struct Root: Decodable {
+		private let items: [Image]
+
+		internal var feedImages: [FeedImage] {
+			return items.compactMap { $0.item }
```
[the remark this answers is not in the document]
fer96: Done

FeedAPIChallenge/Feed API/FeedImageMapper.swift
fer96: Done

FeedAPIChallenge/Feed API/FeedImageMapper.swift
fer96: Nice approach, done 😄

thiagopenna:
Well done!

PR #327, opened 2021-09-10
Valeria Chub - Feed API Challenge

valeriachub:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,17 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		client.get(from: url) { [weak self] result in
+
+			guard self != nil else { return }
+
+			switch result {
+			case let .success((data, response)):
+				completion(FeedItemsMapper.map(data: data, response: response))
+
+			case .failure:
+				completion(.failure(RemoteFeedLoader.Error.connectivity))
```
danillahtin, line 31: `RemoteFeedLoader` can be omitted, it will make the code shorter
```suggestion
				completion(.failure(Error.connectivity))
```

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,51 @@
...
+//  FeedItemsMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Valeria Chub on 10.09.2021.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedItemsMapper {
```
danillahtin, line 11: `internal` is the default access level, so you don't need to write it explicitly.

```suggestion
final class FeedItemsMapper {
```
Suggestion:

`FeedItemsMapper` is defined as a class, so we can instantiate it as `FeedItemsMapper()`. However, all its methods are static - it has no instance methods or properties.

As you'll see in the following lectures, you can make the initializer private in such cases since the `FeedItemsMapper` is just a namespace for static methods, with no instance-specific behavior.

Both solutions will disallow instantiation at compile time so that `FeedItemsMapper()` won't compile ✅
```suggestion
enum FeedItemsMapper {
```

And probably it would be more clear to name this type as FeedImagesMapper as it maps remote images (not items) to local ones.
```suggestion
enum FeedImagesMapper {
```

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,51 @@
...
+
+		return .success(root.feedImages)
+	}
+}
+
+extension FeedItemsMapper {
+	private struct Root: Decodable {
+		private let items: [Item]?
+
+		fileprivate var feedImages: [FeedImage] {
```
danillahtin, line 29: Making this property `fileprivate` has no effect since `Root` is already private in the scope of this file
```suggestion
		var feedImages: [FeedImage] {
```

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
```diff
@@ -0,0 +1,51 @@
...
+		private let url: URL
+
+		private enum CodingKeys: String, CodingKey {
+			case id = "image_id"
+			case description = "image_desc"
+			case location = "image_loc"
+			case url = "image_url"
+		}
+
+		fileprivate var feedImage: FeedImage {
```
danillahtin, line 47: Same here as for `Root.feedImages`
```suggestion
		var feedImage: FeedImage {
```

danillahtin (commented):
Hi @valeriachub, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/FeedItemsMapper.swift
valeriachub: Cool idea about adding the private init() to the class which should not be used for making any instances, as here

danillahtin:
Hi @valeriachub,

PR #328, opened 2021-09-11
Ke Ma - Feed API Challenge.

ankermarco:


FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,41 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Ke Ma on 11/09/2021.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+struct FeedImageMapper {
```
nodediggity, line 11: `FeedImageMapper` is defined as a struct, so we can instantiate it as FeedImageMapper(). However, all its methods are static - it has no instance methods or properties.

It may be confusing for other developers to be able to instantiate an instance but then not be able to do anything with it.

Both solutions will disallow instantiation at compile time so that `FeedImageMapper()` won't compile

```suggestion
enum FeedImageMapper {
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,41 @@
...
+		var feedImage: FeedImage {
+			FeedImage(id: image_id,
+			          description: image_desc,
+			          location: image_loc,
+			          url: image_url)
+		}
+	}
+
+	static func map(data: Data, response: HTTPURLResponse) -> FeedLoader.Result {
+		guard response.statusCode == 200,
```
nodediggity, line 35: It is good practice to avoid using numbers directly in source code without explanation - by defining a named constant in place of numeric literals we can provide more context on what is represented for the next developer.

```swift
```

nodediggity (commented):
Hi @ankermarco,

---

FeedAPIChallenge/Feed API/FeedImageMapper.swift
ankermarco: Thanks a lot Gordon (@nodediggity )

PR #330, opened 2021-09-11
The feed API challenge for Iaroslav Pasternak

Yaro812:


caiozullo:
Hi @Yaro812, please rename the PR following the instructions in the readme file. Thank you!

Yaro812:
Done, thank you!

—

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,30 @@ public final class RemoteFeedLoader: FeedLoader {
...
+				guard response.statusCode == 200 else {
+					return completion(.failure(Error.invalidData))
+				}
+
+				do {
+					let feedImages = try self.feedImages(from: data)
+					completion(.success(feedImages))
+				} catch {
+					completion(.failure(Error.invalidData))
+				}
```
thiagopenna, line 36: Following from my previous comment, by having the `map` function `throw` the correct error we don't have to overwrite it here. That way we could use a `do/catch` statement and complete with the `error` caught.
```swift
	do {
		let feedImages = try self.feedImages(from: data)
	} catch {
		completion(.failure(error))
	}
```

```suggestion
				completion(Result { try self.feedImages(from: data) })
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,30 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+			case .failure:
+				completion(.failure(Error.connectivity))
+			}
+		}
+	}
+
+	private func feedImages(from data: Data) throws -> [FeedImage] {
```
thiagopenna, line 44: We could declare this helper function as `static`, so we won't need a reference to `self` from inside the `load` function.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,30 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard let self = self else { return }
```
thiagopenna, line 23: Following from my previous comment, if we remove the need to upgrade self from a weak reference, we can just check for `nil` instead:
```suggestion
			guard self != nil else { return }
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,30 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+			switch result {
+			case let .success((data, response)):
+				guard response.statusCode == 200 else {
+					return completion(.failure(Error.invalidData))
+				}
```
thiagopenna, line 29: Consider moving this logic to a helper to keep the `load` function shorter and easier to read. Since this portion of code deals with `mapping` the result of `loading`, consider moving this logic to the already existing helper.

thiagopenna (commented):
Hi @Yaro812, you almost passed the challenge!

Thank you 🙂

Yaro812:
Hi Penna,

Thank you for your review. I addressed all the comments. Please consider reviewing one more time

—

Thank you!
Iaroslav Pasternak

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -5,6 +5,10 @@
 import Foundation
 
 public final class RemoteFeedLoader: FeedLoader {
+	enum Constants {
```
thiagopenna, line 8: `Constants` is an implementation detail of `RemoteFeedLoader`, so it should be marked as `private`
```suggestion
	private enum Constants {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -20,29 +24,25 @@ public final class RemoteFeedLoader: FeedLoader {
...
-		let feedResult = try JSONDecoder().decode(FeedResult.self, from: data)
-		return feedResult.items.map { FeedImage($0) }
+	private static func feedLoaderResult(from data: Data, response: HTTPURLResponse) -> FeedLoader.Result {
+		guard
+			response.statusCode == Constants.okCode,
+			let feedResult = try? JSONDecoder().decode(FeedResult.self, from: data) else {
+			return .failure(Error.invalidData)
+		}
+
+		return .success(feedResult.items.map { FeedImage($0) })
```
thiagopenna, line 46: Suggestion: another way of using `map` in this situation is to pass the initializer as a parameter:
```suggestion
		return .success(feedResult.items.map(FeedImage.init))
```

thiagopenna (commented):
Hi @Yaro812,

---

thiagopenna:
Well done!

PR #331, opened 2021-09-11
Omar Abdelmaksoud - Feed API Challenge

omarmohali:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,27 @@ public final class RemoteFeedLoader: FeedLoader {
...
+		}
+	}
+
+	private func handleLoadFeedSuccess(with data: Data, completion: @escaping (FeedLoader.Result) -> Void) {
+		let decoder = JSONDecoder()
+		if let apiResponse = try? decoder.decode(RemoteFeedImages.self, from: data) {
+			completion(.success(RemoteFeedImageToFeedImageMapper.getFeedImages(from: apiResponse.items)))
+		} else {
+			completion(.failure(RemoteFeedLoader.Error.invalidData))
+		}
```
nodediggity, line 43: The `RemoteFeedLoader` should only be responsible for checking the connectivity and passing the additional response values to the `RemoteFeedImageToFeedImageMapper`. 

This allows separation of responsibilities as the Loader implementation only deals with everything related to loading, and the Mapper implementation is responsible for actually mapping the values.

In this case, the loader is handling both loading and verifying the mapped response. 

This can be confusing to another developer as the method `handleLoadFeedSuccess` can complete with a failure, whereas the naming convention implies the operation was already successful.

As we do not know the load operation was successful until we have attempted to decode and map the response, it would perhaps be better to name simply name this `map`.

Additionally, by offloading the mapping work to a static method such as `RemoteFeedImageToFeedImageMapper.map` it is possible to simplify the implementation further by having `load` fail on a connectivity error or deliver the mapping result.

nodediggity (changes requested):
Hi @omarmohali, you almost passed the challenge!

Thank you 🙂

PR #332, opened 2021-09-13
Kaspar Elmans - Feed API Challenge

KasparElmans:
I have implemented the `load` method of the `RemoteFeedLoader`. While also enabling all unit tests.

### DIFF
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -22,23 +22,42 @@ public final class RemoteFeedLoader: FeedLoader {
 		client.get(from: url) { result in
 
 			switch result {
-			case .success(let response):
-				let data = response.0
-				let response = response.1
+			case .success(let (data, response)):
 
 				guard response.statusCode == 200 else {
 					return completion(.failure(Error.invalidData))
 				}
 
-				guard data.count > 0 else {
-					return completion(.success([]))
+				guard let root = try? JSONDecoder().decode(Root.self, from: data) else {
+					return completion(.failure(Error.invalidData))
 				}
+				
+				let feedImages = root.items.map { $0.feedImage }
+				return completion(.success(feedImages))
 
-				// TODO: map data to the model
-				return completion(.success([]))
 			case .failure:
 				return completion(.failure(Error.connectivity))
 			}
 		}
+	}
+}
+
+private struct Root: Decodable {
+	let items: [Item]
+}
+
+private struct Item: Decodable {
+	let id: UUID
+	let description: String?
+	let location: String?
+	let image: URL
+	
+	var feedImage: FeedImage {
+		.init(
+			id: id,
+			description: description,
+			location: location,
+			url: image
+		)
 	}
 }
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -19,7 +19,8 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		client.get(from: url) { result in
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
 
 			switch result {
 			case .success(let (data, response)):
--- a/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
+++ b/FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
@@ -31,7 +31,7 @@ public final class RemoteFeedLoader: FeedLoader {
 				guard let root = try? JSONDecoder().decode(Root.self, from: data) else {
 					return completion(.failure(Error.invalidData))
 				}
-				
+
 				let feedImages = root.items.map { $0.feedImage }
 				return completion(.success(feedImages))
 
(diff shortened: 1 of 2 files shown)


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +20,26 @@ public final class RemoteFeedLoader: FeedLoader {
...
+			switch result {
+			case .success(let (data, response)):
+
+				guard response.statusCode == self.OK_200 else {
+					return completion(.failure(Error.invalidData))
+				}
+
+				guard let root = try? JSONDecoder().decode(Root.self, from: data) else {
+					return completion(.failure(Error.invalidData))
+				}
```
thiagopenna, line 35: Since both `guard` statements complete with the same `Error`, consider joining them:
```suggestion
				guard response.statusCode == self.OK_200, let root = try? JSONDecoder().decode(Root.self, from: data) else {
					return completion(.failure(Error.invalidData))
				}
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +20,26 @@ public final class RemoteFeedLoader: FeedLoader {
...
+
+				let feedImages = root.items.map { $0.feedImage }
+				return completion(.success(feedImages))
```
thiagopenna, line 38: Please extract this `mapping` logic to a helper function, to keep the `load` function shorter and easier to read.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -7,6 +7,7 @@ import Foundation
 public final class RemoteFeedLoader: FeedLoader {
 	private let url: URL
 	private let client: HTTPClient
+	private let OK_200 = 200
```
thiagopenna, line 10: This property could be `static`, so we can access it without an `instance`. Also, there's no reason for each instance of `RemoteFeedLoader` to have it's own value for `OK_200`.
```suggestion
	private static let OK_200 = 200
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +20,26 @@ public final class RemoteFeedLoader: FeedLoader {
...
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard let self = self else { return }
+
+			switch result {
+			case .success(let (data, response)):
+
+				guard response.statusCode == self.OK_200 else {
```
thiagopenna, line 29: Following from my previous comment, by making `OK_200` `static` we won't need to access it through `self`"
```suggestion
				guard response.statusCode == RemoteFeedLoader.OK_200 else {
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +20,26 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard let self = self else { return }
```
thiagopenna, line 24: Following from my previous comments, if we don't need a reference to the current `instance`, there's no need to upgrade it from `weak`. We can just check for `nil` instead:
```suggestion
			guard self != nil else { return }
```

thiagopenna (commented):
Hi @KasparElmans, you almost passed the challenge!

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
KasparElmans: In the future I would like to catch each error separately. But for now there is no use case so joining seems fine 👌🏽

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
KasparElmans: Now self is needed for the `mapRootToFeedImages`. Could extract this method to a class of its own, but this feels like a bit to much at this point.

KasparElmans (commented):
@thiagopenna Thanks for the review! I've added some comment and implemented your feedback! Let me know what you think.

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
thiagopenna: But the `mapRootToFeedImages` does not need to be an `instance` method, it could be marked as `static`. Also, it does not even need to be declared in `RemoteFeedLoader`, and could be part of `Root`, so we might call it as:
```swift
	return completion(.success(root.toFeedImages()))
```

FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
thiagopenna: @KasparElmans, when I said `mapping` logic I was referring to converting from `data` and `response` to `FeedImages`, not just the call to `map`. We can create a helper function like:
```swift
private static func map(_ data: Data, from response: HTTPURLResponse) -> FeedLoader.Result
```

And we can replace the whole `HTTPClient` `success` path with a call to this helper, like:
```swift
switch result {
case .success(let (data, response)):
	return completion(RemoteFeedLoader.map(data, from: response))
case .failure:
	return completion(.failure(Error.connectivity))
}
```

thiagopenna:
> @thiagopenna Thanks for the review! I've added some comment and implemented your feedback! Let me know what you think.

I replied with some extra comments, let me know once you're done so I can review it again.

PR #334, opened 2021-09-14
Daniel Torres - Feed API Challenge

danstorre:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,15 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard self != nil else { return }
+			switch result {
+			case let .success((data, response)):
+				let result = RemoteFeedLoaderParser.map(data: data, response: response)
+				completion(result)
```
knn90, line 27: The `result` variable is not needed, you can inline it.

```suggestion
				completion(RemoteFeedLoaderParser.map(data: data, response: response))
```

FeedAPIChallenge/Feed API/RemoteFeedLoaderParser.swift
```diff
@@ -0,0 +1,46 @@
+
+import Foundation
+
+class RemoteFeedLoaderParser {
```
knn90, line 4: `RemoteFeedLoaderParser` is defined as a class, so we can instantiate it as `RemoteFeedLoaderParser()`. However, all its methods are static - it has no instance methods or properties.

As you'll see in the following lectures, you can make the initializer private in such cases since the `RemoteFeedLoaderParser` is just a namespace for static methods, with no instance-specific behavior.

Both solutions will disallow instantiation at compile time so that `RemoteFeedLoaderParser()` won't compile ✅

FeedAPIChallenge/Feed API/RemoteFeedLoaderParser.swift
```diff
@@ -0,0 +1,46 @@
...
+		}
+
+		init(from decoder: Decoder) throws {
+			let values = try decoder.container(keyedBy: CodingKeys.self)
+			id = try values.decode(UUID.self, forKey: .id)
+			url = try values.decode(URL.self, forKey: .url)
+			description = try values.decodeIfPresent(String.self, forKey: .description)
+			location = try values.decodeIfPresent(String.self, forKey: .location)
+		}
+	}
```
knn90, line 37: The init(from decoder: Decoder) method is not needed here. Since the tests still pass, it's safe to remove the method.

FeedAPIChallenge/Feed API/RemoteFeedLoaderParser.swift
```diff
@@ -0,0 +1,46 @@
...
+
+	static func map(data: Data, response: HTTPURLResponse) -> FeedLoader.Result {
+		guard response.statusCode == 200,
```
knn90, line 40: You can consider creating `private static var OK_200` as we did in the course to give a better context.

knn90 (commented):
Hi @danstorre, you almost passed the challenge!

This is Khoi, I am an instructor at Essential Developer.

Thank you 🙂

FeedAPIChallenge/Feed API/RemoteFeedLoaderParser.swift
danstorre: yeah, because of I didn't have any more codes to be aware of, I just leave it as 200 because it has the context needed.
But I'll added it as requested.

FeedAPIChallenge/Feed API/RemoteFeedLoaderParser.swift
danstorre: yeah great insights, I didn't notice but since is a class it may be inherited as well so I'll add the `final` keyword to the start and also its `internal` access modifier to make explicitly it can only be accessed within the module.

knn90 (commented):
Hi @danstorre,

---

**P.S. I left a final review comment. Feel free to push changes.**

PR #335, opened 2021-09-15
 Mohamed Ibrahim - Feed API Challenge.

MIF50:


FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,41 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by MIF50 on 15/09/2021.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+final class FeedImageMapper {
```
danillahtin, line 11: `FeedImageMapper` is defined as a class, so we can instantiate it as `FeedImageMapper()`. However, all its methods are static - it has no instance methods or properties.

```suggestion
enum FeedImageMapper {
```

FeedAPIChallenge/Feed API/FeedImageMapper.swift
```diff
@@ -0,0 +1,41 @@
...
+	private struct Root: Decodable {
+		let items: [Item]
+		var feeds: [FeedImage] {
```
danillahtin, line 14: In the context of the app a `feed` is a collection of feed images (`FeedImage`). Thus feeds should mean a collection of collections of feed images. Thus this property should be either `feed` or `images`

Please make the property name match with its type
```suggestion
		var feed: [FeedImage] {
```

danillahtin (commented):
Hi @MIF50, you almost passed the challenge!

Thank you 🙂

PR #336, opened 2021-09-15
Hardeep Singh - Feed API Challenge

hardeep-singh:


FeedAPIChallenge/Feed API/RemoteFeedLoader.swift
```diff
@@ -19,6 +19,16 @@ public final class RemoteFeedLoader: FeedLoader {
 	}
 
 	public func load(completion: @escaping (FeedLoader.Result) -> Void) {
-		fatalError("Must be implemented")
+		client.get(from: url) { [weak self] result in
+			guard self != nil else {
+				return
+			}
+			switch result {
+			case .success((let data, let httpResponse)):
```
danillahtin, line 27: Suggestion:
The `let` can be defined outside the tuple so the expression can be shorter:
```suggestion
			case let .success((data, httpResponse)):
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,45 @@
...
+//  FeedImageMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Hardeep Singh on 15/09/21.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+final class FeedImagesMapper {
```
danillahtin, line 11: `FeedImagesMapper` is defined as a class, so we can instantiate it as `FeedImagesMapper()`. However, all its methods are static - it has no instance methods or properties.

```suggestion
enum FeedImagesMapper {
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,45 @@
...
+		var image: FeedImage? {
+			guard let imageURL = URL(string: image_url) else {
+				return nil
+			}
+			return FeedImage(id: image_id,
+			                 description: image_desc,
+			                 location: image_loc,
+			                 url: imageURL)
+		}
+	}
```
danillahtin, line 34: `URL` [conforms](https://github.com/apple/swift/blob/d1f798bedb61d4c137573360e6e623670efedc2a/stdlib/public/Darwin/Foundation/URL.swift#L1201-L1227) to `Codable`, there's no need to create it manually from string.
```suggestion
	private struct Item: Codable {
		let image_id: UUID
		let image_url: URL
		let image_loc: String?
		let image_desc: String?

		var image: FeedImage? {
			return FeedImage(id: image_id,
			                 description: image_desc,
			                 location: image_loc,
			                 url: image_url)
		}
	}
```

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,45 @@
...
+
+	private static var OK_200: Int { 200 }
```
danillahtin, line 36: It is absolutely ok to use implicit returns but they are not used in this project. It is important to be consistent and keep nice and concise code style. Please make the return explicit
```suggestion
	private static var OK_200: Int { return 200 }
```

danillahtin (commented):
Hi @hardeep-singh, you almost passed the challenge!

Thank you 🙂

PR #339, opened 2021-09-20
Vadims Vorobjovs - Feed API Challenge

VadVorobjov:


FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,44 @@
...
+//  FeedImagesMapper.swift
+//  FeedAPIChallenge
+//
+//  Created by Vadims Vorobjovs on 15/09/2021.
+//  Copyright © 2021 Essential Developer Ltd. All rights reserved.
+//
+
+import Foundation
+
+internal final class FeedImagesMapper {
```
knn90, line 11: `internal` is the default access level, so you don't need to write it explicitly.

knn90: `FeedImageMapper` is defined as a class, so we can instantiate it as `FeedImageMapper()`. However, all its methods are static - it has no instance methods or properties.

FeedAPIChallenge/Feed API/FeedImagesMapper.swift
```diff
@@ -0,0 +1,44 @@
...
+			return FeedImage(id: image_id,
+			                 description: image_desc,
+			                 location: image_loc,
+			                 url: image_url)
+		}
+	}
+
+	private static var OK_200 = 200
+
+	internal static func map(_ data: Data, from response: HTTPURLResponse) -> RemoteFeedLoader.Result {
```
knn90, line 36: Same as the previous comment, please remove the `internal` keyword.

knn90 (commented):
Hi @VadVorobjov, you almost passed the challenge!

This is Khoi, I am an instructor at Essential Developer.

Thank you 🙂

knn90:
Hi @VadVorobjov,

VadVorobjov:
Hello Khoi,

Thank you for your review!

Best wishes,
Vadims

